PR #4, opened 2023-06-12
Fetch user journey events from Splunk

ifireball:
A set of shell scripts for:
- Fetching RHTAP UserJounrney events from Splunk
- Cross-referencing usernmaes in events with UserSignup resources to map from cluster username for SSO user IDs
- Formatting the events into Segment event records suitable for uploading via the bulk send API.

### DIFF
--- a/scripts/fetch-uj-records.sh
+++ b/scripts/fetch-uj-records.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# fetch-uj-records.sh
+#   Fetch user journey records from Splunk
+#   Records look almost like Segment track records but some further adjustments
+#   are required such as:
+#   - Converting nested JSON objects from strings to objects
+#   - Mapping cluster usernames to SSO user IDs
+#   This script assumes that credentials are preconfigured for curl for
+#   connecting to Splunk in a .netrc file
+#
+set -o pipefail -o errexit -o nounset
+
+SPLUNK_API_URL=https://splunk-api.corp.redhat.com:8089
+SPLUNK_APP_NAME=rh_rhtap
+SPLUNK_APP_API_URL="$SPLUNK_API_URL/servicesNS/nobody/$SPLUNK_APP_NAME"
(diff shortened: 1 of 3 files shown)


scripts/fetch-uj-records.sh
```diff
@@ -0,0 +1,55 @@
...
+    \"namespace\", 'objectRef.namespace',
+    \"name\", 'objectRef.name'
+  )
+  | fields messageId, timestamp, type, userId, event_verb, event_subject, properties | fields - _*
+"
+
+QUERY="$SQ_EV_SELECTOR | $SQ_DEDUP_FIELDS | $SQ_GEN_JSON"
+
+set -o xtrace
+curl -k -n \
```
Omeramsc, line 52: Might be worthwhile mentioning this script assumes the username and password (with the right permissions) are configured

scripts/splunk-to-segment.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+# splunk-to-segment.sh
+#   Adapt user journey recordes loaded from Splunk for uploading into Segmment:
+#   - Map cluster usernames to SSO user IDs
+#   - Convert nested JSON objects from strings to actual objects.
```
Omeramsc, line 5: This is mentioned on the first script as well, is the conversion done twice? I don't see this script using `fetch-uj-records.sh`, or the opposite

Omeramsc (changes requested):

scripts/splunk-to-segment.sh
ifireball: This script implements the things that `fecth-uj-records.sh` specifies that need to be implemented, the scripts do not use each other, they are meant to be piped together.

scripts/fetch-uj-records.sh
ifireball: done

Omeramsc (approved):
LGTM

PR #5, opened 2023-06-18
Limit Splunk query to last 4 hours

ifireball:
Change `fetch-uj-records.sh` so that if only retrives records from the last 4 hours by default.

Also made some other consmetic changes:
- Moved all the script variables that may need to be customized for testing purposes or manually resolving issues to one documented spot.
- Made the query return operations made directly by users, not just ones that were done their behalf by system accounts.
- Removed the namespace field from the query results as that may expose PII to Segment.

### DIFF
--- a/scripts/fetch-uj-records.sh
+++ b/scripts/fetch-uj-records.sh
@@ -9,40 +9,55 @@
 #   connecting to Splunk in a .netrc file
 #
 set -o pipefail -o errexit -o nounset
+#
+# ======= Parameters ======
+# The following variables can be set from outside the script by setting
+# similarly named environment variables.
+# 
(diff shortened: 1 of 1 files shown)


Omeramsc (approved):
LGTM

yftacherzog:
I'd change the PR/commit title. Looking at the implementation, accomplishing what the title implies would be a 5-lines PR, but most of this PR is about something else, which could not have been guessed reading the title.

ifireball:
> I'd change the PR/commit title. Looking at the implementation, accomplishing what the title implies would be a 5-lines PR, but most of this PR is about something else, which could not have been guessed reading the title.

updated

yftacherzog (approved):

PR #6, opened 2023-06-18
chore(RHTAPWATCH-283): Containerized Splunk for tests

Omeramsc:
Configures a Splunk instance with prebuilt index and indexed logs for future testing purposes.


Omeramsc:
@ifireball  PTAL
also a good opportunity to mention I cannot assign reviewers :)

splunk/tests/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY test_logs /opt/splunk/test_logs
+COPY indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
+COPY ../scripts/index_verification.sh /opt/splunk/index_verification.sh
+
+RUN microdnf install -y jq && \
+    microdnf clean all
+
+RUN /sbin/entrypoint.sh start-service &\
+    sleep 60 &&\
```
ifireball, line 19: Maybe loop and curl the service instead of blindly waiting a minute.

splunk/scripts/index_verification.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+logs_path="$SPLUNK_HOME/test_logs"
+endpoint="https://localhost:8089/services/data/inputs/oneshot"
+
+echo "Indexing logs..."
+for file in "$logs_path"/*; do
+  echo "Indexing $file"
+  curl -k -u admin:"$SPLUNK_PASSWORD" "$endpoint" -d name="$file"
```
ifireball, line 9: Do we really need -k here? The image if missing the certificate for a a service that is running inside itself? Also prefer longer option names for better readability in scripts.

splunk/scripts/index_verification.sh
```diff
@@ -0,0 +1,20 @@
...
+done
+
+response=$(curl -k -u admin:"$SPLUNK_PASSWORD" "$endpoint" -X GET -d output_mode=json)
+indexing_status=$(echo "$response" | jq -r '.paging.total')
```
ifireball, line 13: Can use `--exit-status` to do the boolean check from inside `jq` so you can do this whole thing in a single `if` statement instead of 3 different statements. Also no reason to store the `curl` response instead of piping in directly to `jq`, this is just allocating an unnecessary string.

splunk/scripts/index_verification.sh
```diff
@@ -0,0 +1,20 @@
...
+logs_path="$SPLUNK_HOME/test_logs"
+endpoint="https://localhost:8089/services/data/inputs/oneshot"
+
+echo "Indexing logs..."
+for file in "$logs_path"/*; do
+  echo "Indexing $file"
+  curl -k -u admin:"$SPLUNK_PASSWORD" "$endpoint" -d name="$file"
+done
+
+response=$(curl -k -u admin:"$SPLUNK_PASSWORD" "$endpoint" -X GET -d output_mode=json)
```
ifireball, line 12: Can use `--get` as a higher-level alternative to `-X GET`

splunk/tests/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM docker.io/splunk/splunk:latest
+
+ENV SPLUNK_START_ARGS=--accept-license
+ENV SPLUNK_PASSWORD=$SPLUNK_PASSWORD
```
ifireball, line 4: Perhaps set a temporary password in the indexing script instead a fixed one in the image that you need to configure at build time.

ifireball (changes requested):

splunk/tests/Dockerfile
Omeramsc: SPLUNK_PASSWORD must be setup on the image, otherwise the service will fail setting up

ifireball: It aught to be sufficient to have it be set in the environment while running `/sbin/entrypoint.sh start-service, I don't see anything else that may be using it, and we may want a different password once we run the image we build as part of a test.

splunk/tests/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
+COPY ../scripts/index_verification.sh /opt/splunk/index_verification.sh
+
+RUN microdnf install -y jq && \
+    microdnf clean all
+
+RUN /sbin/entrypoint.sh start-service &\
+    sleep 60 &&\
+    chmod +x /opt/splunk/index_verification.sh &&\
+    /opt/splunk/index_verification.sh
```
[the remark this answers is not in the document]
Omeramsc: I didn't forget, that's actually the issue I'm facing and why running the container itself fails - I couldn't get the service pid or the service to shutdown. commands like ps yield nothing...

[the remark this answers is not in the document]
ifireball: Well since its a blocking command, you'll have at least the PID of the parent script in `$!`

splunk/scripts/index_verification.sh
ifireball: On SSL errors?

splunk/tests/Dockerfile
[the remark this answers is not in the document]
Omeramsc: Even so, it at least seems like the service is still open / the ports are open, blocking the runtime service from working.

splunk/scripts/index_verification.sh
Omeramsc: Yes

splunk/tests/Dockerfile
Omeramsc: done

splunk/scripts/index_verification.sh
Omeramsc: done

splunk/scripts/index_verification.sh
Omeramsc: done

splunk/tests/Dockerfile
Omeramsc: Will do

splunk/tests/test_logs/some_log.log
```diff
@@ -0,0 +1,1320 @@
+127.0.0.1 - omer-system-user [10/Jun/2023:10:11:55.504 +0000] "GET //services/server/info?output_mode=json HTTP/1.0" 200 2155 "-" "Python-httplib2/0.13.1 (gzip)" - - - 13ms
```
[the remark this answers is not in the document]
Omeramsc: Will do, currently I just tried putting something long with an identifier of mine so I'll be able to assure it is working

splunk/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY scripts/index_verification.sh /opt/splunk/index_verification.sh
+
+RUN microdnf install -y jq && \
+    microdnf clean all
+
+USER ansible
+
+RUN SPLUNK_PASSWORD="Password" /sbin/entrypoint.sh start-service &\
+    sleep 60 &&\
+    bash /opt/splunk/index_verification.sh
```
ifireball, line 21: so is that working now without clean shutdown?

Omeramsc: The index verification itself is working, but running the container fails because AFAIU, the Splunk service is already open. I cannot seem to shut it off

Omeramsc:
```
FAILED - RETRYING: Start Splunk via CLI (5 retries left).
FAILED - RETRYING: Start Splunk via CLI (4 retries left).
FAILED - RETRYING: Start Splunk via CLI (3 retries left).
FAILED - RETRYING: Start Splunk via CLI (2 retries left).
FAILED - RETRYING: Start Splunk via CLI (1 retries left).

TASK [splunk_common : Start Splunk via CLI] ************************************
fatal: [localhost]: FAILED! => {
    "attempts": 5,
    "changed": false,
    "cmd": [
        "/opt/splunk/bin/splunk",
        "start",
        "--accept-license",
        "--answer-yes",
        "--no-prompt"
    ],
    "delta": "0:00:01.044448",
    "end": "2023-06-19 14:25:09.996017",
    "rc": 1,
    "start": "2023-06-19 14:25:08.951569"
}

STDOUT:


Splunk> See your world.  Maybe wish you hadn't.

Checking prerequisites...
	Checking http port [8000]: open
	Checking mgmt port [8089]: open
	Checking appserver port [127.0.0.1:8065]: open
	Checking kvstore port [8191]: open
	Checking configuration... Done.


STDERR:

The CA file specified (/opt/splunk/etc/auth/cacert.pem) does not exist. Cannot continue.
SSL certificate generation failed.


MSG:

non-zero return code
```

ifireball: > The index verification itself is working, but running the container fails because AFAIU, the Splunk service is already open. I cannot seem to shut it off

The service cannot possibly stay open between build time and when you run the container - those are two different containers with different process spaces and network stacks.

As far as the log you quote goes - it seems to me that starting the service causes it to generate certificates for itself that are not getting cleaned up of saved properly because we're not shutting it down in the most orderly way.

What we should probably do is clean the directories back to how they were before we started it before completing the build. One thing you can try is to save a recursive scan of the filenames in `/opt/splunk/etc/` (With E.g. `find`), before and after you run the service and then compare the two file lists.

Also some nitpicks:

- `/opt/splunk/index_verification.sh` is not a good name for what it is doing
- Not sure why you keep living some things out of the script instead of having one organized script to do the whole indexing chore.

Omeramsc: I went over all the file additions (seems like there are only new files, no changes), over a thousand, removed them all during the build and... still having the same issue. 

as for the nitpicks, - the name is not final and will change, and it is is one script rn, just did not update the PR yet as I'm trying to solve the main issue...

Omeramsc:
Updated with the latest changes, still does not work properly. 
main changes:
1. now works as a multi-stage build
2. the certificates does not exist before running the splunk command. they get created afterwards. I left the part of the script that removes the certificates as well as the one which removes all file changes in the splunk directory. unfortunately, it still fails with the same error.

splunk/scripts/pre-run.sh
```diff
@@ -0,0 +1,26 @@
...
+/sbin/entrypoint.sh start-service &
+sleep 60
+
+# Verifying the indexing of the logs
+bash /opt/splunk/index_verification.sh
+
+# Option 1: removing only the auth directory with the cacert.pem file
+rm -r /opt/splunk/etc/auth
+
+# Option 2: removing all file additions
```
ifireball, line 17: no need for this anymore because of multistage

splunk/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY scripts/index_verification.sh /opt/splunk/index_verification.sh
+COPY scripts/pre-run.sh /opt/splunk/pre-run.sh
+
+RUN microdnf install -y jq && \
+    microdnf clean all
+
+# USER ansible    # root user is still needed for removing files
+RUN bash /opt/splunk/pre-run.sh
+
+FROM builder AS splunk
```
ifireball, line 21: the whole point of doing multistage here is to start the final image from a clean base so that the sins you did in the build stage no longer hunt you. Doing "FROM builder" beats this purpose....

splunk/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+COPY tests/test_logs /opt/splunk/test_logs
+COPY tests/indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
+COPY scripts/index_verification.sh /opt/splunk/index_verification.sh
+COPY scripts/pre-run.sh /opt/splunk/pre-run.sh
+
+RUN microdnf install -y jq && \
+    microdnf clean all
+
+# USER ansible    # root user is still needed for removing files
```
ifireball, line 18: you have to make sure you never try to run splunk as root, because it wasn't meant to run like that

ifireball (changes requested):

splunk/Dockerfile
Omeramsc: as long as we don't need to remove the files, no issues accomplishing that. will fix

splunk/Dockerfile
Omeramsc: Not sure I understand you then. which image should the second one start from, splunk as well? we decided on performing such "sins" because we need to assure we have the logs indexed before running any test on the container...

ifireball: It should state from the same clean image that you started the build phase from.
And yes you should copy the files, with the usually copy commands between stages.

The whole point of using multistage here is to do this:
1. Start from a clean image
2. Mess it up with some process (E.g. indexing)
3. Remove all the mess by restarting from clean image again
4. Copy just the files we need from the messed up image so we end up with a clean image

ifireball: Frankly if you didn't understand this, why were you trying to use multi-stage in the first place? just because I said so?

Omeramsc: I did understand what you meant when you suggested it, but did not fully understand the implementation itself. 
--
as for copying the files - I'm not sure which of the 1500+ files changed are the one relevant for us, 
but just to try, I copied the whole Splunk directory. funny enough, it works (I get the splunk instance running), but even with everything copied, the indexing seem to revert..

Omeramsc: This means only the some_log.log file, which is the one I'm copying to the container in the first place. I'm pretty sure this file is left unchanged (I checked), so hopefully we are not reversing the indexing... (for now it seems like either we are, or the indexing never happend)

ifireball: See `indexes.conf`, you list 3 files there:
```
coldPath = $SPLUNK_DB/$_index_name/colddb 
 homePath = $SPLUNK_DB/$_index_name/db 
 thawedPath = $SPLUNK_DB/$_index_name/thaweddb
```
Seems likely to me those are the files the index is stored in...

Omeramsc: I thought so too, but $SPLUNK_HOME/var/lib/splunk (which is the default for $SPLUNK_DB) does not exist, and $SPLUNK_DB leads to nowhere... I couldn't find the `$_index_name` directory at all.

ifireball: try to see what happens wen you run the image and create the index normally. These variables may be set by the startup script and not be preset in the image

splunk/tests/Dockerfile
Omeramsc: Done

splunk/Dockerfile
Omeramsc: Done, it is finally working. 
Building & Running the container now should get us a Splunk container with some logs attached to `test_index` index.

splunk/scripts/pre-run.sh
Omeramsc: done

splunk/Dockerfile
Omeramsc: Done, it is finally working.
Building & Running the container now should get us a Splunk container with some logs attached to test_index index.

splunk/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+ENV SPLUNK_START_ARGS=--accept-license
+
+EXPOSE 8080
+EXPOSE 8089
+
+USER root
+
+RUN mkdir -p /opt/splunk/var/lib/splunk
+COPY tests/test_logs /opt/splunk/test_logs
```
ifireball, line 28: we probably don't need to log files in the image because they are already indexed

splunk/scripts/pre-run.sh
```diff
@@ -0,0 +1,26 @@
...
+    fi
+    if [ "$(($(date +%s) - timeout_start))" -ge 120 ]; then
+        echo "The Splunk instance is not up, cannot verify indexing for tests." >&2
+        exit 1
+    fi
+    sleep 5
+done
+
+# Verify the indexing of the logs
+bash /opt/splunk/log_indexing.sh
```
ifireball, line 24: that seems to actually do the indexing, not just verify it

splunk/scripts/pre-run.sh
```diff
@@ -0,0 +1,26 @@
...
+
+export SPLUNK_PASSWORD="Password"
+
+# Running an instance of Splunk during the build preparing the indexed logs for the tests
+/sbin/entrypoint.sh start-service &
+
+# Make sure the Splunk service is up
+timeout_start=$(date +%s)
+while true; do
+    response=$(curl --insecure -s -o /dev/null -w "%{http_code}" "https://localhost:8089/")
```
ifireball, line 11: use long option names

splunk/scripts/log_indexing.sh
```diff
@@ -0,0 +1,18 @@
...
+endpoint="https://localhost:8089/services/data/inputs/oneshot"
+SPLUNK_PASSWORD="Password"
+
+echo "Indexing logs..."
+for file in "$logs_path"/*; do
+  echo "Indexing $file"
+  curl  --insecure --user admin:"$SPLUNK_PASSWORD" "$endpoint" --data name="$file" --data index="test_index"
+done
+
+if [[ $(curl --insecure --user admin:"$SPLUNK_PASSWORD" "$endpoint" --get --data output_mode=json | jq -r --exit-status 'if .paging.total == 0 then 0 else 1 end') != 0 ]]; then
```
ifireball, line 13: breaking lines for readability is not illegal...

Also:
```
if [[ $(foo) != 0 ]]; then ...
```
checks that `foo` did not **print** 0, to check it did not **return** 0 you need to:
```
if foo; then ...
```

You already used `--exit-status` for `jq` which means it would return `0` if its internal calculation results in `true`, but you're not using that.

splunk/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN bash /opt/splunk/pre-run.sh
+
+FROM docker.io/splunk/splunk:latest
+
+ENV SPLUNK_START_ARGS=--accept-license
+
+EXPOSE 8080
+EXPOSE 8089
+
+USER root
```
ifireball, line 25: you need to remember to change this back so the final image does not run as root.

also if you `COPY` as root you end up with root-owned files so you need to either properly chmod and chown them (IIRC there are flags to `COPY` to do that), or not `COPY` as  root to begin with.

splunk/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM docker.io/splunk/splunk:latest AS builder
```
ifireball, line 1: lets use a fixed label of SHA so upstream image updates will not break our CI.

ifireball (changes requested):

splunk/scripts/log_indexing.sh
Omeramsc: I tried a different variation with ```if ! curl --insecure --user admin:"$SPLUNK_PASSWORD" "$endpoint" --get --data output_mode=json \
  | jq -e '.paging.total == 0'
  ```
  but it didn't work as expected. will see why and fix.

ifireball:
waiting for all other comments to be resolved

splunk/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+ENV SPLUNK_START_ARGS=--accept-license
+
+EXPOSE 8080
+EXPOSE 8089
+
+USER root
+
+RUN mkdir -p /opt/splunk/var/lib/splunk
+COPY --chmod=777 --from=builder /opt/splunk/etc/apps/search/local/indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
```
ifireball, line 30: 777? seriously? it should at least be 666 or 665 or probably 660 or even 600

splunk/scripts/log_indexing.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Indexing logs..."
+for file in "$logs_path"/*; do
+  echo "Indexing $file"
+  curl  --insecure --user admin:"$SPLUNK_PASSWORD" "$endpoint" --data name="$file" --data index="test_index"
+done
+
+timeout_start=$(date +%s)
+while true; do
+    if ! curl --insecure --user admin:"$SPLUNK_PASSWORD" "$endpoint" --get --data output_mode=json | jq -e '.paging.total == 0'; then
```
ifireball, line 15: newlines are good for your soul... For example the `;` between `if` and `then` could be a newline...

splunk/scripts/log_indexing.sh
```diff
@@ -0,0 +1,26 @@
...
+        echo "Waiting for log indexing to complete." >&2
+        break
+    fi
+    if [ "$(($(date +%s) - timeout_start))" -ge 60 ]; then
```
ifireball, line 19: '[ .. ]' runs a 'test' subprocess and is generally based on a very old and quirky syntax. Use `[[ ... ]]`

ifireball (changes requested):
one security issue and a couple of stylistic issues

splunk/Dockerfile
Omeramsc: Anything with less permissions than 777 leads to errors during runtime (Ansible playbook fails).

splunk/scripts/log_indexing.sh
Omeramsc: fixed

splunk/scripts/log_indexing.sh
Omeramsc: fixed

ifireball (approved):

PR #8, opened 2023-06-20
Update README with current architecture

ifireball:
Updated the README file to include details about our current shell-script based architecture.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,43 +3,47 @@
 Bridge selected events from AppStudio into [Segment][1]
 
 ```mermaid
-flowchart TD
+flowchart TB
     subgraph A["RHTAP clusters"]
         A1[(API server)]
         A2(["OpenShift
-         logging collector"])
+        logging collector"])
 
         A1--"Audit Logs"-->A2
     end
 
-    A2--"Audit Logs"--->C1
+    A2--"Audit Logs"-->C
     
-    subgraph C["Splunk"]
-        C1[("Splunk index")]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,43 +3,47 @@
...
 
 ```mermaid
-flowchart TD
+flowchart TB
     subgraph A["RHTAP clusters"]
         A1[(API server)]
-        A2(["OpenShift
-         logging collector"])
+        A2(["`OpenShift
+        logging collector`"])
```
yftacherzog, line 10: What's the meaning/purpose of adding the ` accent?

README.md
```diff
@@ -3,43 +3,47 @@
...
+        B3--"Segment events"--> B4
     end
 
-    C3--"User resource
-     actions"-->B2
+    C-- "User resource
+     actions" -->B2
 
     G([Segment])
     H[(Aplitude)]
```
yftacherzog, line 43: is that a typo? Can we fix it while we're at it? 
```suggestion
    H[(Amplitude)]
```

README.md
ifireball: It makes the text be processed as markdown, but unfortunately this is not supported by the Mermaid version used by GitHub, so removed.

README.md
ifireball: fixed.

yftacherzog (approved):

Omeramsc (approved):
LGTM

PR #9, opened 2023-06-29
fix(RHTAPWATCH-355): Remove false creation events

ifireball:
We found out that sometimes we get two resource create events in the audit log, both of which are accepted by the API server but only one of which is actually creating the resource record.

To tell which is the actual creation record we need to ensure we have a `resourceVersion` field in the `respobseObject` we get back from the creation action.


Omeramsc (approved):
LGTM

PR #10, opened 2023-07-02
feat(RHTAPWATCH-300): Add container image

ifireball:
Adding a container image to run the UJ data sync pipeline.

To make everything runnable in a container, made the location of the `~/.netrc` file configuratble via an enviromnet variable, so it can be made to point to a file mounted on the container.

PR #11, opened 2023-07-04
feat(RHTAPWATCH-310): Add dockerfile for kwok

srivickynesh:
Adding a container image bring up a KWOK container. 
We should be able to run queries with oc get.
JIRA:[ RHTAPWATCH-310 ](https://issues.redhat.com/browse/RHTAPWATCH-310)


kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Start from a base image
+FROM registry.access.redhat.com/ubi8:latest
```
ifireball, line 2: and we use an existing kwok image as base?

kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Update packages
+RUN yum update -y && yum clean all
```
ifireball, line 5: Never do this! it makes you image build unreproducible!

kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN yum install -y bash
```
ifireball, line 7: No way you need this, bash is built-in to the image

Also because this is a sperate RUN step - you leave the yum cache in the image layer.

kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install curl and other dependencies
+RUN yum install -y curl tar && yum clean all
```
ifireball, line 10: all those tools are probably already in the UBI image as well.

kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install Kwok & Kwokctl
+RUN KWOK_VERSION=$(curl -s https://api.github.com/repos/kubernetes-sigs/kwok/releases/latest | grep 'tag_name' | cut -d\" -f4) && \
```
ifireball, line 13: `curl`ing during a build is a bad idea - it will not work in RHTAP. 

Either:
1. Use a base image that already contains kwok
2. Use the golang tools to vendor the kwok source code as a dependency (I assume its written in Go) and build it during the container build.

kwok/commands.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+kwokctl create cluster --name=kwok
+kwokctl get clusters
+kubectl cluster-info --context kwok-kwok
+oc get namespace
```
ifireball, line 5: Why do we need to do this as part of the image build?

kwok/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    curl -LO https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERSION}/kwok-linux-amd64 && \
+    chmod +x kwok-linux-amd64 && \
+    mv kwok-linux-amd64 /usr/local/bin/kwok && \
+    curl -LO https://github.com/kubernetes-sigs/kwok/releases/download/${KWOK_VERSION}/kwokctl-linux-amd64 && \
+    chmod +x kwokctl-linux-amd64 && \
+    mv kwokctl-linux-amd64 /usr/local/bin/kwokctl
+
+ARG OCP_VERSION=4.12.23
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${OCP_VERSION}/openshift-client-linux-${OCP_VERSION}.tar.gz | tar xvz -C /usr/local/bin/
```
ifireball, line 22: Can use that same base image I've used in #10 which already contains `oc` though I'm not sure we really need it.

kwok/commands.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
ifireball, line 1: I'd rather you find a way to do the cluster creation and image build time and not at image start time.

ifireball (changes requested):

kwok/commands.sh
srivickynesh: Im just showcasing here the oc cmd works. that's also part of AC.

ifireball: yeah, but a showcase should not become a permanent part of the image...

kwok/Dockerfile
srivickynesh: Don't we require control over oc version?

kwok/Dockerfile
srivickynesh: could you add the kwok base image here?

ifireball: You are telling me you couldn't find such an image?

kwok/Dockerfile
srivickynesh: I tried the second approach there were some dependencies that is unresolved. 
[root@5d851c9cbae6 /]# kwok
kwok: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by kwok)
kwok: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by kwok)
[root@5d851c9cbae6 /]# kwokctl
kwokctl: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by kwokctl)
kwokctl: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by kwokctl)

kwok/Dockerfile
srivickynesh: we might need Kwok base image in Quay. since docker images have a rate limit. we might have to find something that resides with our goal.

kwok/Dockerfile
ifireball: it seems you need to install the devel version of glibc to build kwok.

did you use the go toolset image to build?

ifireball: Actually, looking again at the erorr - it seems the library is missing at runtime - are you using the same RHEL/UBI version to build and run?

kwok/Dockerfile
ifireball: not sure we'll build frequently enough to hit that rate limit

kwok/Dockerfile
srivickynesh: go-toolset latest id 1.19 and kwok requires 1.20 to install via go mod. so did a workaround to install via debian

kwok/commands.sh
srivickynesh: removed

kwok/Dockerfile
srivickynesh: removed

kwok/commands.sh
srivickynesh: Done

kwok/Dockerfile
srivickynesh: removed

kwok/Dockerfile
srivickynesh: removed

kwok/Dockerfile
ifireball: so that is causing your problem because Debian's libc is very different from Red Hat's - you need to find a way to only use Red Hat stuff - perhaps we can sun some command in the toolset to update go there somehow

srivickynesh: updated to use v0.2.0

kwok/Dockerfile
srivickynesh: Updated

kwok/Dockerfile
srivickynesh: Done

Omeramsc (approved):
Overall LGTM, that being said I prefer another set of eyes on this as I'm not aware of how kwok works or what are its dependencies.

kwok/Dockerfile
ifireball: not sure I follow, are you referring to the kwok version?

srivickynesh: yep

README.md
```diff
@@ -154,6 +154,14 @@ then 4 times in a row and issue an appropriate alert.
 
 ## Running a test environment
 
+### Kwok Docker Container
+
+This repository contains a Dockerfile to build a container for the Kwok Kubernetes tool.
+
+1. `cd kwok`
```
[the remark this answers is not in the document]
srivickynesh: Updated

kwok/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN go install sigs.k8s.io/kwok/cmd/kwok@v0.3.0
+RUN go install sigs.k8s.io/kwok/cmd/kwokctl@v0.2.1
+
+FROM registry.access.redhat.com/ubi8/ubi:latest
+
+COPY --from=builder /opt/app-root/src/go/bin/kwok /usr/local/bin/
+COPY --from=builder /opt/app-root/src/go/bin/kwokctl /usr/local/bin/
+
+RUN command -v kwok && command -v kwokctl
```
ifireball, line 19: I guess this is for testing? Probably shouldn't stay in the final version...

Also what about setting up a cluster inside the image?

README.md
```diff
@@ -154,6 +154,14 @@ then 4 times in a row and issue an appropriate alert.
...
 ## Running a test environment
 
+### Kwok Docker Container
+
+This repository contains a Dockerfile to build a container for the Kwok Kubernetes tool. Navigate to the 'kwok' directory, build the Docker container for the Kwok Kubernetes tool using the provided Dockerfile, and then run the created Docker image interactively.
+
+1. `cd kwok`
+2. `docker build -t kwok-image .`
+3. `docker run -it --rm kwok-image:latest`
+
```
ifireball, line 164: Missing motes about how to use the image as a cluster

ifireball (changes requested):

README.md
```diff
@@ -154,6 +154,14 @@ then 4 times in a row and issue an appropriate alert.
 
 ## Running a test environment
 
+### Kwok Docker Container
```
eisraeli, line 157: Why build it from scratch when we can use a pre-built image?
https://kwok.sigs.k8s.io/docs/user/all-in-one-image

kwok/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:1.19 AS builder
```
eisraeli, line 1: See my comment on the README.md file about using a pre-built image.

README.md
eisraeli: Overall, I feel this PR should have been about how to USE the KWOK image to function as a pseudo-cluster rather than BUILDING KWOK - as KWOK is already being offered as a binary and as a pre-built image.

srivickynesh: Done

README.md
srivickynesh: Done

README.md
ifireball: We need a way to connect to the cluster in the container from outside the container. The given `exec` command only does that from inside the container.

Also, can you please use podman instead of docker?

ifireball (changes requested):

README.md
srivickynesh: Updated

srivickynesh: The image doesn't have bash or sh. another way to run outside the container is to add the Kube config and export it.

srivickynesh: or we may have to revert to use ubi image. @ifireball  wdyt?

ifireball: what does running outside of the container have to do with having a shell inside the container?

we need a way to setup a predictable port and predictable credentials for the cluster in the container so that we know how to connect to it from kubectl/oc running outside.

srivickynesh: In the updated step we can access using the cluster name and port is also exposed.

ifireball: Does `kwokctl get clusters` print how to connect to the cluster? 

Consider that what we are going to have in the end is a test in Go that is going to bring up the kwok image, run our {{splunk-to-segment.sh}} script while providing a kube config and then inspect the output. Its preferable of we don't have to run any special kwok commands in order to figure out how to connect to it.

amisstea: @ifireball it does but this should also be printed [when the container starts](https://github.com/kubernetes-sigs/kwok/blob/90f923e7c8320d79e98b5cf3dbd0edfce35a1fa7/images/cluster/entrypoint.sh#L27-L39). It's mostly static so we can probably commit this kubeconfig. The port is configurable and defaults to [8080](https://github.com/kubernetes-sigs/kwok/blob/90f923e7c8320d79e98b5cf3dbd0edfce35a1fa7/images/cluster/Dockerfile#L21).

@srivickynesh we should show a sample `kubectl` command using such a kubeconfig.

ifireball: @amisstea ok.

@srivickynesh can we get this done by end of sprint?

README.md
```diff
@@ -154,6 +154,28 @@ then 4 times in a row and issue an appropriate alert.
...
+
+1. Bring up the Kwok cluster by running the following command:
+
+`podman run --rm -d -p 8080:8080 --name kwok-cluster registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3`
+
+2. Once the Kwok cluster is up and running, set the cluster details in the OpenShift client with the following command:
+
+`oc config set-cluster kwok --server=http://127.0.0.1:8080`
+
+Now we can access the cluster using kubectl cmds.
```
ifireball: this sentence probably appears too soon....

README.md
```diff
@@ -154,6 +154,35 @@ then 4 times in a row and issue an appropriate alert.
...
+```
+podman run --rm -d -p 8080:8080 --name kwok-cluster registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
+```
+
+2. Once the Kwok cluster is up and running, set the cluster details in the OpenShift client with the following command:
+
+```
+oc config set-cluster kwok --server=http://127.0.0.1:8080
+```
+   Now we can access the cluster using kubectl cmds.
```
ifireball, line 172: this sentence probably appears too soon...

ifireball (changes requested):

README.md
```diff
@@ -154,6 +154,35 @@ then 4 times in a row and issue an appropriate alert.
...
+```
+oc config set-context kwok --cluster=kwok
+```
+
+4. Set the Kwok context as the current context with the following command:
+
+```
+oc config use-context kwok
+```
+
```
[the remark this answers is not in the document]
srivickynesh: done

README.md
srivickynesh: once we set the cluster it also sets the context from kubeconfig. no need to change it manually

kwok/Dockerfile
srivickynesh: Done

README.md
srivickynesh: Updated

README.md
```diff
@@ -154,6 +154,34 @@ then 4 times in a row and issue an appropriate alert.
 
 ## Running a test environment
 
+### Kwok Container
+
```
yftacherzog, line 158: I think it's worth adding a sentence here about what is Kwok.

README.md
```diff
@@ -154,6 +154,34 @@ then 4 times in a row and issue an appropriate alert.
...
+```
+podman run --rm -d -p 8080:8080 --name kwok-cluster registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
+```
+
+2. Once the Kwok cluster is up and running, set the cluster details in the OpenShift client with the following command:
+
+```
+oc config set-cluster kwok --server=http://127.0.0.1:8080
+```
+   Now you can access the cluster using kubectl, for eg: `kubectl get ns`.
```
yftacherzog, line 172:
```suggestion
   Now you can access the cluster using kubectl, e.g.: `kubectl get ns`.
```

srivickynesh: Done

README.md
srivickynesh: Done

ifireball:
@yftacherzog can you re-review please? Since I had a part in coming up with the solutions here, I may know too much to realize wither we explain things clearly or not.

yftacherzog (approved):

hmariset:
LGTM

PR #12, opened 2023-07-05
Sekelton Go files

ifireball:
It took a while to figure out where to put things.

This is not 100% done, I've no idea yet how to build and run things from the CLI, without `cd`ing all over the place, but at least the VSCode Go plugin can use this layout nicely.

### DIFF
--- a/cmd/querygen/main.go
+++ b/cmd/querygen/main.go
@@ -1,0 +1,11 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/redhat-appstudio/segment-bridge.git/querygen"
+)
+
+func main() {
+	fmt.Println(querygen.GenQuery())
+}
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/redhat-appstudio/segment-bridge.git
+
+go 1.19
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -1,0 +1,5 @@
+package querygen
+
+func GenQuery() string {
+	return "Hi!"
+}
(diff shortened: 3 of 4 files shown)


Omeramsc (approved):
nice, LGTM, 

Do you mean running `go build` without cd'ing? not sure it is possible.

unless  we'll use Go mod to to set the querygen as a go module so we can run it from root

ifireball:
> nice, LGTM,
> 
> Do you mean running `go build` without cd'ing? not sure it is possible.
> 
> unless we'll use Go mod to to set the querygen as a go module so we can run it from root

I did use Go mod - the entire repo is setup as a module - but that did not seem to do the trick here. I really want to be able to use `go` as a `pipenv` replacement - but it looks like we have a learning curve to get there

Omeramsc:
> > nice, LGTM,
> > Do you mean running `go build` without cd'ing? not sure it is possible.
> > unless we'll use Go mod to to set the querygen as a go module so we can run it from root
> 
> I did use Go mod - the entire repo is setup as a module - but that did not seem to do the trick here. I really want to be able to use `go` as a `pipenv` replacement - but it looks like we have a learning curve to get there

From what I read, the alternative is the set the `querygen` directory as a module as well as the root directory... but I'm not sure that's the intended usage. Then, you can use go build from the root directory to build the root package as well as the package within the `querygen` directory

ifireball:
> From what I read, the alternative is the set the `querygen` directory as a module as well as the root directory... but I'm not sure that's the intended usage. Then, you can use go build from the root directory to build the root package as well as the package within the `querygen` directory

Well this does not really add up for me - to the best of my understanding - querygen is a package, and a module is supposed to be able to contain multiple packages...

I really miss `__init__.py` files, having those really made things clear IMO...

PR #13, opened 2023-07-06
refactor(RHTAPWATCH-116): `querygen` Go program

ifireball:
Create the `querygen` Go program for generating query strings for Splunk
and use that in `fetch-uj-records.sh` instead of the bash code we had
for this before.

Doing this as a first step to switching to having multiple queryes and
more complex ones. Using Go for this allows for more complex query
generating logic then wnat we can easily do in `bash`.

This PR depends on #12

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -1,5 +1,123 @@
+// Package querygen is used to generate Splunk queries for fetching user journey
+// events from the RHTAP K8s event log.
 package querygen
 
-func GenQuery() string {
-	return "Hi!"
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+const (
+	// UserIdExpr is a Splunk eval expression for obtaining the human username
+	// behind a given event even if the event action was done on their behalf by
+	// a service account
+	UserIdExpr = `if(isnull('impersonatedUser.username'),` +
+		`'user.username','impersonatedUser.username')`
+
+	// IncludeFieldsCmd is a Splunk query "fields" command for specifying which
+	// fields to include in the final query results. Since the structure of a
+	// Segment record is quite fixed, most queries should return the same
+	// fields.
+	IncludeFieldsCmd = `fields ` +
+		`messageId, timestamp, type, userId, event_verb, event_subject, properties`
+
+	// ExcludeFieldsCmd is a Splunk query "fields" command for removing fields
+	// from query results that splunk includes by default and we don't need.
+	ExcludeFieldsCmd = `fields - _*`
+)
+
+// GenDedupEval generates a Splunk "eval" command for converting multivalue
+// fields into single-value fields. See RHTAPWATCH-293 for why we need this.
+func GenDedupEval(fields []string) string {
+	var builder strings.Builder
+	builder.WriteString("eval ")
+	for i, field := range fields {
+		if i > 0 {
+			builder.WriteString(",")
+		}
+		fmt.Fprintf(&builder, `%s=mvindex('%s', 0)`, field, field)
+	}
+	return builder.String()
+}
+
+// GenPropertiesJSONExpr generates a Splunk "eval" expression for generating a
+// JSON object containing the fields and values specified in properties_map. The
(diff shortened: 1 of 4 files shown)


querygen/querygen.go
```diff
@@ -0,0 +1,96 @@
...
+
+var (
+	UserIdExpr = `if(isnull('impersonatedUser.username'),` +
+		`'user.username','impersonatedUser.username')`
+	IncludeFieldsCmd = `fields ` +
+		`messageId, timestamp, type, userId, event_verb, event_subject, properties`
+	ExcludeFieldsCmd = `fields - _*`
+)
+
+func GenDedupEval(fields []string) string {
```
Omeramsc, line 17: I think [Doc-comments](https://go.dev/doc/comment)  (=GO docstrings) will improve readability of this package (applicable in multiple parts of the file)

querygen/querygen_test.go
```diff
@@ -0,0 +1,151 @@
...
+		name string
+		args args
+		want string
+	}{
+		{
+			name: "with userid, verb, subject & app props",
+			args: args{
+				spec: TrackFieldSpec{
+					with_userid:     true,
+					with_ev_verb:    true,
```
Omeramsc, line 128: We have a parameterize test, but don't we have only the case in which all 3 booleans are true?

Omeramsc (changes requested):

querygen/querygen_test.go
ifireball: well, I was putting this in place for the time I add an actual query with different values, but since it took long enough to get to where we are, I pushed that off to the future.

querygen/querygen.go
ifireball: added a bunch of comments.

querygen/querygen_test.go
ifireball: can we resolve this?

Omeramsc (approved):
LGTM

hmariset (approved):
LGTM

querygen/querygen.go
```diff
@@ -1,5 +1,123 @@
...
-	return "Hi!"
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+const (
+	// UserIdExpr is a Splunk evel expression for otaining the human username
+	// behind a given event even if the even action was done on their behalf by
```
amisstea, line 13:
```suggestion
	// UserIdExpr is a Splunk eval expression for obtaining the human username
	// behind a given event even if the event action was done on their behalf by
```

querygen/querygen.go
```diff
@@ -1,5 +1,123 @@
...
+		`'user.username','impersonatedUser.username')`
+
+	// IncludeFieldsCmd is a Splunk query "fields" command for specifying which
+	// fields to include in the final query results. Since the structure of a
+	// Segment record is quite fixed, most queries should return the same
+	// fields.
+	IncludeFieldsCmd = `fields ` +
+		`messageId, timestamp, type, userId, event_verb, event_subject, properties`
+
+	// IncludeFieldsCmd is a Splunk query "fields" command for removing fields
```
amisstea, line 25:
```suggestion
	// ExcludeFieldsCmd is a Splunk query "fields" command for removing fields
```

querygen/querygen.go
```diff
@@ -1,5 +1,123 @@
...
+	for i, field := range fields {
+		if i > 0 {
+			builder.WriteString(",")
+		}
+		fmt.Fprintf(&builder, `%s=mvindex('%s', 0)`, field, field)
+	}
+	return builder.String()
+}
+
+// GenPropertiesJSONExpr generates a Splunk "evel" expression for generating a
```
amisstea, line 44:
```suggestion
// GenPropertiesJSONExpr generates a Splunk "eval" expression for generating a
```

amisstea (approved):
There are just a few small typos in the comments. LGTM otherwise.

srivickynesh (approved):

PR #17, opened 2023-07-16
feat(RHTAPWATCH-371): add script mapping namespace to workspace

yftacherzog:
For each cluster, the script searches for namespaces with label toolchain.dev.openshift.com/type set to tenant and maps the namespace name to the value of label toolchain.dev.openshift.com/space.

### DIFF
--- a/scripts/get-workspace-map.sh
+++ b/scripts/get-workspace-map.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# get-workspace-map.sh
+#   Read Namespace resources and generate a JSON object mapping from namespaces
+#   to workspace based on the value of label toolchain.dev.openshift.com/space.
+#   This script assumes `oc` is preconfigured with all the required clusters
+#   and could be found in $PATH.
+#
+#   if the CONTEXTS environment variable is set, the script will query
+#   locally-defined contexts with those names. Otherwise, the script will query
+#   contexts associated with a hard-coded list of staging clusters.
+#
+set -o pipefail -o errexit -o nounset
+
+# URLs for staging member clusters to be queried if contexts were not provided
+# explicitly. A local context with the same URL must exist for each element below (can be
+# named arbitrarily).
+# I.e. for each url u, kubeconfig must have an entry c on the list named contexts for
+# which c.context.cluster == u
+stage_clusters=$(cat <<- EOM
+  [
+    "api-stone-stg-m01-7ayg-p1-openshiftapps-com:6443",
+    "api-stone-stg-rh01-l2vh-p1-openshiftapps-com:6443"
+  ]
+EOM
+)
+
+# ======= Parameters ======
+# The following variables can be set from outside the script by setting
+# similarly named environment variables.
+#
+# Locally-defined context names to be queried (space-separated)
+read -r -a CONTEXTS <<< "${CONTEXTS:-""}"
+#
+# === End of parameters ===
+
+if [[ ${#CONTEXTS[@]} -eq 0 ]]; then
+  mapfile -t CONTEXTS < <(
+    kubectl config view -o=json \
+    | jq --argjson clusters "$stage_clusters" \
+    '.contexts | .[] | select(.context.cluster as $c | $clusters | index($c)) | .name'
+  )
+fi
+
+printf "%s\n" "${CONTEXTS[@]}" | xargs -r --replace=C \
+  oc --context=C get namespaces \
+    --selector=toolchain.dev.openshift.com/type=tenant \
+    -o=go-template \
+    --template=$'{
+      {{- $comma := false}}
+      {{- range .items}}
+          {{- $namespace := .metadata.name }}
+          {{- $workspace := index .metadata.labels "toolchain.dev.openshift.com/space" }}
+          {{- if $comma}},{{end}}
+          {{- $comma = true -}}
+          "{{$namespace}}":"{{ $workspace }}"
+      {{- end -}}}\n' \
+  | jq --slurp 'add'


scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,30 @@
...
+#   to workspace based on the value of label toolchain.dev.openshift.com/space.
+#   This script assumes `oc` is preconfigured with all the required clusters
+#   and could be found in $PATH
+#
+set -o pipefail -o errexit -o nounset
+
+out={}
+for context in "$@"
+do
+  oc config use-context "$context"
```
ifireball, line 13: This is prone to race conditions, perhaps we can use the `--context` parameter instead? Or maybe there is some env var we can use?

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,30 @@
...
+# get-workspace-map.sh context-1 context-2... context-n
+#   Read Namespace resources and generate a JSON object mapping from namespaces
+#   to workspace based on the value of label toolchain.dev.openshift.com/space.
+#   This script assumes `oc` is preconfigured with all the required clusters
+#   and could be found in $PATH
+#
+set -o pipefail -o errexit -o nounset
+
+out={}
+for context in "$@"
```
ifireball, line 11: Id prefer we use an env var with some useful default value rather than requiring parameters to be explicitly set in each call

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,30 @@
...
+        {{- $comma := false}}
+        {{- range .items}}
+            {{- $namespace := .metadata.name }}
+            {{- $workspace := index .metadata.labels "toolchain.dev.openshift.com/space" }}
+            {{- if $comma}},{{end}}
+            {{- $comma = true -}}
+            "{{$namespace}}":"{{ $workspace }}"
+        {{- end -}}}\n'
+  )
+  out=$(echo "$out" "$new"  | jq -s '.[0] * .[1]')
```
ifireball, line 28: I think there is a way to stream all the different cluster maps into jq at one and have it merge them all instead of running a different jq command for each pair (IIRC it has at least a `reduce` function)

This will also avoid relying on shell buffers which are prone to be limited.

eisraeli, line 28: Using 
```
out=$(jq -s '.[0] * .[1]' <<< "$out $new")
```
will not create a sub-shell, thus it might be better.

ifireball: Looking at the jq docs: while we have reduce, the add function may be sufficient in this case

scripts/get-workspace-map.sh
eisraeli: Can't we do something like `oc get namespaces --context="$context"`

yftacherzog: I'll give it a try. I was not aware of that option.

ifireball:
Given the `oc` `--prefix` option, the `xargs` `--delimiter` option and the `jq` `add` command and `--slurp` option, this script can probably be reduced to a single shell pipeline:
```
xargs <<<"$CLUSTER_PREFIXES" --delimitr=: --replace=C oc --prefix=C get ... | jq --slurp "add(.)"
```

Where CLUSTER_PREFIXES is a colon-delimeted list of prefixes, though using space as a delimiter may be better in this case.

scripts/get-workspace-map.sh
yftacherzog: What might be a good default value for that?

scripts/get-workspace-map.sh
yftacherzog: done

yftacherzog:
> Given the `oc` `--prefix` option, the `xargs` `--delimiter` option and the `jq` `add` command and `--slurp` option, this script can probably be reduced to a single shell pipeline:
> 
> ```
> xargs <<<"$CLUSTER_PREFIXES" --delimitr=: --replace=C oc --prefix=C get ... | jq --slurp "add(.)"
> ```
> 
> Where CLUSTER_PREFIXES is a colon-delimeted list of prefixes, though using space as a delimiter may be better in this case.

@ifireball Can you point me to documentation for the `--prefix` option? I can find some mentions of it, but not to the extent that I could tell what it does.

scripts/get-workspace-map.sh
ifireball: What I'd really love to have happen here is that if you clone the repo and run the script locally it just works - assuming you already have permissions to the clusters somehow. How close can we get to that? Can we somehow make is so the context definitions come from some file in the repo while the connection credentials come from the user's `~/.kube/config` file? As an alternative, perhaps we can leverage openid?

ifireball:
> 
> @ifireball Can you point me to documentation for the `--prefix` option? I can find some mentions of it, but not to the extent that I could tell what it does.

s/prefix/context/  :facepalm:

scripts/get-workspace-map.sh
yftacherzog: What if we assume that the context is defined and we have the script grep for lines that has the cluster name in it?

e.g. if we run it for stage, we use the user's pre-stored contexts that include `stg-m01` or `stg-rh01`. This way, we don't have to know the exact name on one hand (for me it's stored with a name that includes my github username), and we don't have to know how to fetch it on the other hand.

ifireball: Given that `.kube/config` is a YAML file - `grep` would not be the right tool for this, `yq` would, but that's adding another dependency. Maybe if we grep on `oc config get-contexts`.

Ultimately this feels like a bit of an overkill that is not reusable for the "running as an automated job" use case. I would prefer to have a pre-backed kubeconfig file to neutralize any garbage that may be in the user's file.

I really feel like OpenID with the `oidc-login` plugin may work - are you familiar with how that is used to connect to our systems?

yftacherzog: (I didn't mean using the grep toll necessarily - just to rely on the contexts that the user has)

I'm not familiar with anything related to OpenID other than the fact that I click on it as authentication option from time to time.

yftacherzog: I feel we have two different tasks here that make each other more complicated.

What if we make the script to accept the contexts as it is at the moment, and we later on have it run another script to fetch the contexts in case contexts are not provided?

We can then implement the process of fetching the contexts independently of the current issue, and that process can also be used for other scripts to require something similar.

ifireball: Have a look here: 
https://docs.google.com/document/d/1hFvQDH1H6MGNqTGfcZpyl2h8OIaynP8sokZohCS0Su0/edit#heading=h.olhho0rpp8t5

This document actually complicates things because of the need to use the RHTAP proxy. Using the `kubectl` [oidc-login][1] plug-in is even simpler. Once you have things set up you get to the point where using a properly configured context just works, or otherwise pops up your browser to login to the cluster if you're logged out.

If you've never installed a kubectl/oc plug-in before (Any kubectl plug-in also works with oc), I recommend using [krew][2] for this. You can obtain krew itself [via asdf][3] if you're using it.

[1]: https://github.com/int128/kubelogin
[2]: https://krew.sigs.k8s.io/
[3]: https://github.com/nlamirault/asdf-krew

ifireball: I don't want to defer the problem of figuring out which contexts to use - its an inherent part of writing this script (Regardless of wither you factor it out to a different script or not). I don't foresee other scripts having to solve a similar problem for now.

ifireball: So a quick attempt at setting things up via OIDC failed. I added this context to my configuration:
```
- context:
    cluster: api-stone-stg-m01-7ayg-p1-openshiftapps-com:6443
    namespace: default
    user: oidc
  name: rhtap-stage-m01
```

But trying to use that gave me: 
```
error: You must be logged in to the server (Unauthorized)
```

This is despite something very similar working well via the RHTAP proxy (Effectively going to the same cluster).

Here is the setup for the oidc user, the `rhoas-cli-prod` thing makes me suspect there is more here then just going via `sso.redhat.com` which all clusters do.
```
- name: oidc
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - oidc-login
      - get-token
      - --oidc-issuer-url=https://sso.redhat.com/auth/realms/redhat-external
      - --oidc-client-id=rhoas-cli-prod
      command: oc
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

So I suppose its back to grepping the user's existing configuration.

yftacherzog: Any way we turn it, I think one thing is missing from the script at the moment is being able to handle failures. As I see it, the script needs to replace the output from clusters it failed fetching with `{}` and to somehow report that at the end.

ifireball: I think we might get that for free, if `oc` fails to connect it would dump some error into STDERR and nothing to STDOUT, so we'll end up with `{}` from `jq` - but you probably should test that assumption.

scripts/get-workspace-map.sh
yftacherzog: done

scripts/get-workspace-map.sh
yftacherzog: This is done now. Probably there are some ways to make it prettier, but let me know if you're ok with the approach.

yftacherzog: @ifireball This is done now. Probably there are some ways to make it prettier, but let me know if you're ok with the approach.

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,92 @@
...
+
+if [[ -z $STAGE && -z $PROD && -z $EXPLICIT ]]; then
+  echo "No clusters found."
+  exit 1
+fi
+
+contexts="${STAGE} ${PROD} ${EXPLICIT}"
+
+out={}
+for context in $contexts
```
ifireball, line 73: are you planning to get rid of this `for` in favour of `xargs` and a single pipeline?

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,92 @@
...
+    echo "No production clusters found."
+  fi
+fi
+
+if [[ -z $STAGE && -z $PROD && -z $EXPLICIT ]]; then
+  echo "No clusters found."
+  exit 1
+fi
+
+contexts="${STAGE} ${PROD} ${EXPLICIT}"
```
ifireball, line 70: I don't think we need to support a combination of stage, prod and explicitly requested. IMO defaulting to stage would be enough for a reasonable local dev experience.

What I'm aiming for is something similar to what we did in other scripts where we have an env var that would be used when running in a POD, with a default value set to it to make local test runs easy.

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,92 @@
...
+  )
+  if [[ -z $STAGE ]]; then
+    echo "No staging clusters found."
+  fi
+fi
+
+if [[ -n $PROD ]]; then
+# Take all non-host production clusters and keep only their names
+  PROD=$(
+    oc config get-contexts \
```
ifireball, line 53: all the `grep`s and `sed`s can probably be combined to a single `sed` run. For example here is how you implement a simple grep with sed:
```
sed -n '/regex-here/p'
```
Where `p` is a simple print command but there are other commands such as `d` that can be used to filter out lines, and the well-known `s`.

I don't think the `tr` is needed either because the following would make base get rid on the newlines for us:
```
FOO=($(foo | bar))

File: scripts/get-workspace-map.sh (hunk 4 above)
Comment (yftacherzog) [reply scripts/get-workspace-map.sh:73]: I'm trying. I wasn't able to make it work yet.

File: scripts/get-workspace-map.sh (hunk 5 above)
Comment (yftacherzog) [reply scripts/get-workspace-map.sh]: I think it will be easier to generalize the support the combination vs. just one at each time.

Anyhow, I implemented the use of env var instead of script arguments with querying stage as default behavior.

File: scripts/get-workspace-map.sh (hunk 6 above)
Comment (yftacherzog) [reply scripts/get-workspace-map.sh:42]: I'll give it a try. I don't like it when there's too much magic involved. Hard to troubleshoot.

File: scripts/get-workspace-map.sh (hunk 4 above)
Comment (yftacherzog) [reply scripts/get-workspace-map.sh:62]: What I'm unable to do, at least not in a pretty way, and for something which isn't much simpler than what we're trying to do here, is to combine at output of all the `oc` executions to one `jq` invocation.

File: scripts/get-workspace-map.sh [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+  echo "No clusters found."
+  exit 1
+fi
+
+contexts="${STAGE} ${PROD} ${QUERY_URLS}"
+
+out={}
+for context in $contexts
+do
+  echo "Querying cluster with context: $context..."
```
amisstea, line 62: The script currently returns malformed JSON due to this line. Please remove it.

yftacherzog: oh yes

yftacherzog: done

scripts/get-workspace-map.sh
yftacherzog: I've been looking into replacing those multiple `grep` and `sed` calls to a single `sed` or even at replacing parts of it. I was not able to get it to work yet, but I think the result would be less maintainable than the current approach.

regex and sed already have the potential of becoming write-only code. I'd prefer leaving it as a sequence of simple steps rather than a single harder-to-manage step.

ifireball: I do want to find a way to simplify this - I think this logic for getting the clusters is occupying too much code ATM - perhaps we can leave only QUERY_URLS and drop the rest while making it default to the stage hosts?

Also I think we need a way to hard-wire the `contexts` variable for the automated use case where we control the contents of the kubeconfig file.

yftacherzog: > I do want to find a way to simplify this - I think this logic for getting the clusters is occupying too much code ATM - perhaps we can leave only QUERY_URLS and drop the rest while making it default to the stage hosts?

Not sure what you mean.

> Also I think we need a way to hard-wire the contexts variable for the automated use case where we control the contents of the kubeconfig file.

you mean that by default the script will use all contexts in the kubeconfig file?

yftacherzog: Turns out there's a `kubectl config view` that can be processed with `-o=go-template`

ifireball: nice! 

lets setup a few minutes to meet and iron out the behaviour around this

yftacherzog: @ifireball please take another look.

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+#   if the CONTEXTS environment variable is set, the script will query
+#   locally-defined contexts with those names. Otherwise, the script will query
+#   contexts associated with a hard-coded list of staging clusters.
+#
+set -o pipefail -o errexit -o nounset
+
+# Space-separated locally-defined context names to be queried
+CONTEXTS=(${CONTEXTS:-""})
+stage_clusters=$(cat <<- EOM
```
Omeramsc, line 16: since this is used only in the conditional statement, any reason not to set it there?

yftacherzog: Yes. I wanted it to be seen as a constant. Suppose that tomorrow the list of staging clusters changes, I want this constant to be at the top of the file so it's easy to locate and update.

scripts/get-workspace-map.sh
```diff
@@ -0,0 +1,45 @@
...
+#   and could be found in $PATH.
+#
+#   if the CONTEXTS environment variable is set, the script will query
+#   locally-defined contexts with those names. Otherwise, the script will query
+#   contexts associated with a hard-coded list of staging clusters.
+#
+set -o pipefail -o errexit -o nounset
+
+# Space-separated locally-defined context names to be queried
+CONTEXTS=(${CONTEXTS:-""})
```
ifireball, line 15: Lets document this like we document the parameters for other scripts

ifireball (approved):
just a little nitpick

scripts/get-workspace-map.sh
yftacherzog: done

yftacherzog: I now changed the way the contexts env var is expanded and evaluated. It should be safer now, and should now also attend to the case in which somehow the first provided context is an empty string.

https://github.com/redhat-appstudio/segment-bridge/compare/a73b70c6517b28af9283f3a6634cfa80e37d4f0d..07d6a565fe32595cfde8ad45607cb51e4c617ea3

ifireball (approved):

scripts/get-workspace-map.sh
yftacherzog: Also changed the way the array of contexts is being split from the jq command, so ShellCheck don't complain.

https://github.com/redhat-appstudio/segment-bridge/compare/07d6a565fe32595cfde8ad45607cb51e4c617ea3..533ff4f7b86c5e0ecdd2c5234fe756940d9d455d

Omeramsc (approved):
LGTM

PR #18, opened 2023-07-18
test(RHTAPWATCH-311): Mini HTTP test framework

ifireball:
Adding a little test function that would be useful for testing things that make simple calls into web servers, and do not care too much about the response. The main candidate for sing this is
 `segment-mass-uploader.sh`.

### DIFF
--- a/webfixture/webfixture.go
+++ b/webfixture/webfixture.go
@@ -1,0 +1,38 @@
+// Package webfixture includes a simple HTTP-server test fixture that collects
+// the requests made to it
+package webfixture
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+)
+
+// RequestTrace represents simple data about a web request
+type RequestTrace struct {
+	Method,	Path, Body string
+}
+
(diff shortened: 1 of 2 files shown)


webfixture/webfixture.go
```diff
@@ -0,0 +1,38 @@
...
+// RequestTrace represents simple data about a web request
+type RequestTrace struct {
+	Method,	Path, Body string
+}
+
+// TraceRequestsFrom runs a small web server, and then invokes a provided test
+// function while passing it the server URL and an HTTP client. The requests
+// made to the web server while the test function is running are then logged
+// and returned.
+func TraceRequestsFrom(test_func func(url string, c http.Client)) (requests []RequestTrace) {
```
srivickynesh, line 20:
```suggestion
func TraceRequestsFrom(test_func func(url string, c *http.Client)) (requests []RequestTrace) {
```

srivickynesh (approved):
LGTM

Omeramsc (approved):
LGTM
ice use of goroutine and naked return

PR #19, opened 2023-07-20
feat(RHTAPWATCH-323): add usersignup to kwok cluster

srivickynesh:
Jira: [RHTAPWATCH-323](https://issues.redhat.com/browse/RHTAPWATCH-323) 
Use base image to create kwok cluster and usersignup script to create usersignups

cmd to test this :
`podman build --build-arg CLUSTER_NAME=tester -t custom-image .`
`podman run --rm -it -p 8080:8080 custom-image`

### DIFF
--- a/kwok/Dockerfile
+++ b/kwok/Dockerfile
@@ -1,0 +1,14 @@
+FROM registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
+
+ARG CLUSTER_NAME=kwok
+ENV CLUSTER_NAME=${CLUSTER_NAME}
+
+COPY --chmod=755 user-signups.sh entrypoint.sh toolchain.dev.openshift.com_usersignups.yaml ./
+
+RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
+    kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    sh ./user-signups.sh && \
+    rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
+    kwokctl stop cluster --name="$CLUSTER_NAME"
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/kwok/entrypoint.sh
+++ b/kwok/entrypoint.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+kwokctl start cluster --name "$CLUSTER_NAME"
+kwokctl --name "$CLUSTER_NAME" kubectl proxy --port="${KWOK_KUBE_APISERVER_PORT}" --accept-hosts='^*$' --address="0.0.0.0"
--- a/kwok/toolchain.dev.openshift.com_usersignups.yaml
+++ b/kwok/toolchain.dev.openshift.com_usersignups.yaml
@@ -1,0 +1,166 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.10.0
+  creationTimestamp: null
+  name: usersignups.toolchain.dev.openshift.com
+spec:
+  group: toolchain.dev.openshift.com
+  names:
+    kind: UserSignup
+    listKind: UserSignupList
+    plural: usersignups
+    singular: usersignup
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .spec.username
+      name: Username
+      type: string
+    - jsonPath: .spec.givenName
+      name: First Name
+      priority: 1
+      type: string
+    - jsonPath: .spec.familyName
+      name: Last Name
+      priority: 1
+      type: string
+    - jsonPath: .spec.company
+      name: Company
+      priority: 1
+      type: string
+    - jsonPath: .spec.targetCluster
+      name: TargetCluster
+      priority: 1
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="Complete")].status
+      name: Complete
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="Complete")].reason
+      name: Reason
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="Approved")].status
+      name: Approved
+      priority: 1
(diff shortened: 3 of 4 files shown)


kwok/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
+
+COPY user-signups.sh ./user-signups.sh 
+
+RUN chmod +x ./user-signups.sh
+
+COPY entrypoint.sh ./entrypoint.sh
+
+RUN wget https://raw.githubusercontent.com/codeready-toolchain/host-operator/master/config/crd/bases/toolchain.dev.openshift.com_usersignups.yaml
```
ifireball, line 9: please don't curl/wget doing image build - either copy that file to our repo or link it via git submodules.

kwok/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN chmod +x ./user-signups.sh ./entrypoint.sh ./toolchain.dev.openshift.com_usersignups.yaml
```
ifireball, line 11: you can chmod with the copy command, no need to create an extra image layer for that.

kwok/entrypoint.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo "# The above example works if your host's port is the same as the container's,"
+  echo "# otherwise, change it to your host's port"
+}
+
+# Create a cluster
+KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster "$@" || exit 1
+
+show_info &
+
+kwokctl kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && sh ./user-signups.sh
```
ifireball, line 56: since you are running user-signups.sh from the entry point script you are essentially loading the data during image startup instead of during the build. We want to do this during the build.

kwok/entrypoint.sh
```diff
@@ -0,0 +1,56 @@
...
+
+  echo "###############################################################################"
+  echo "# The above example works if your host's port is the same as the container's,"
+  echo "# otherwise, change it to your host's port"
+}
+
+# Create a cluster
+KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster "$@" || exit 1
+
+show_info &
```
ifireball, line 54: we don't need this show-info thing in our custom script

ifireball (changes requested):

kwok/entrypoint.sh
srivickynesh: we don't have a preexisting cluster from the image. the cluster itself is created during run time, applying this during the build process won't work

kwok/Dockerfile
srivickynesh: Done

kwok/Dockerfile
srivickynesh: Done

kwok/entrypoint.sh
srivickynesh: Done

kwok/entrypoint.sh
ifireball: My aim is to create the cluster at build time. I can see `kwok` as `start` and `stop` commands, so we can probably create at build time, stop and then start it back when the image starts.

srivickynesh: Done

Omeramsc (approved):
LGTM

kwok/entrypoint.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+kwokctl start cluster "$@"
```
ifireball, line 2: in the original dockerfile I can see that entrypoint scrip also brings up a proxy to listen on  the API port and forward requests to the cluster. Are you sure we don't need that? Did you try connecting to this image from outside?

kwok/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
+
+COPY --chmod=755 user-signups.sh entrypoint.sh toolchain.dev.openshift.com_usersignups.yaml ./
+
+RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster "$@" || exit 1 && \
```
ifireball, line 5: we probablyy don't need to use `$@` here - I guess you copied this command from the original entry point script, but this is meant to allow passing parameters to the image to customize the cluster name and other elements, we don't need that and it can't work that way during the image build.

kwok/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+    kwokctl kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    sh ./user-signups.sh && \
```
ifireball, line 7: instead of uploading a script like this and running it doing the build, we can just the script once and commit the results to the repo as static YAML files. That way we don't have an unnecessary script left behind in the image and we can more easily customize the records later.

kwok/toolchain.dev.openshift.com_usersignups.yaml
```diff
@@ -0,0 +1,166 @@
+---
```
ifireball, line 1: While having a copy of the file makes things simple for now, we need to think how do we keep it up to date.

ifireball (changes requested):

kwok/entrypoint.sh
srivickynesh: Since you have mentioned removing it from the entry point I have removed it

kwok/toolchain.dev.openshift.com_usersignups.yaml
srivickynesh: Keeping this yaml up to date can be automated.

kwok/entrypoint.sh
ifireball: I mentioned removing the part that prints out the connection details, not the part that brings up the proxy. If you are able to connect to the cluster without it then by all means, lets keep things simple. By the way how to we set the port for the API server - we need to be able to customize that at image load time to avoid colliding with other ports on the user's machine.

srivickynesh: Done

kwok/Dockerfile
srivickynesh: Done

ifireball (approved):

amisstea (approved):

PR #20, opened 2023-07-20
chore(RHTAPWATCH-361): GO test framework

Omeramsc:
Adds the containerfixture script that will be served as a go test framework, with the ability to build and deploy containerized services for us to perform tests on with real data.
    
The Splunk container specifically is injected with indexes and logs for the tests, therefor requiring specific verification before running the tests.

Currently the script intentionaly does not use the podman/bindings library in attempt to avoid adding too many uneeded dependencies.


splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,73 @@
...
+func RunSplunk() {
+	filePath := "runtime.txt"
+	file, err := os.Create(filePath)
+	if err != nil {
+		log.Fatalf("Failed to create file: %v", err)
+	}
+	defer file.Close()
+
+	cmd := exec.Command("podman", "run", "-e", "SPLUNK_START_ARGS=--accept-license", "-e", "SPLUNK_PASSWORD=Password", "-p", "8000:8000", "-p", "8089:8089", "localhost/test_splunk:latest")
+	cmd.Stdout = file
```
ifireball, line 35: I don't think we need to route the command output to a file like that - we can see the container output with `podman logs` if we need to.

We can let any stray lines go the the usual stdout I think, especially if we'll use the `-d` option for `podman run`. The default handling of stderr by `cmd.Run` or `cmd.Output` seems sufficient to me.

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,73 @@
...
+	buildCmd.Stderr = os.Stderr
+	err := buildCmd.Run()
+	if err != nil {
+		log.Fatalf("Failed to build image: %v", err)
+	}
+	log.Println("Image build done successfully!")
+}
+
+// RunSplunk deploy the Splunk container used for testing Splunk queries.
+func RunSplunk() {
```
ifireball, line 26: We probably need a way to stop it too... I wonder if `go test` has a way to have session-wide fixtures

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,73 @@
...
+	filePath := "runtime.txt"
+	file, err := os.Create(filePath)
+	if err != nil {
+		log.Fatalf("Failed to create file: %v", err)
+	}
+	defer file.Close()
+
+	cmd := exec.Command("podman", "run", "-e", "SPLUNK_START_ARGS=--accept-license", "-e", "SPLUNK_PASSWORD=Password", "-p", "8000:8000", "-p", "8089:8089", "localhost/test_splunk:latest")
```
ifireball, line 34: We probably can't assume the image is there aleady, perhaps we should use the `kube play` thing which can also build the images as needed.

splunk/tests/splunk_test_prep.go
ifireball: Ok, I see you are using the file as a monitoring device.... Why use a file when you can use a simple descriptor instead? 

The problem with this approach is that the podman process remains attached to our Go process and we must make sure to maintain it and clean up after it. It would probably be better to use `-d` and monitor the service network port as a way of knowing if its up - monitoring the port would also be what we need when we run in a cluster.

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,73 @@
...
+
+	err = cmd.Start()
+	if err != nil {
+		log.Fatalf("Failed to start command: %v", err)
+	}
+	serviceAvailable := VerifyServiceIsUp(filePath)
+	if !serviceAvailable {
+		log.Fatalf("Failed to run container: %v", err)
+	}
+	log.Println("Splunk service deployed successfully!")
```
ifireball, line 46: You never wait for the process to complete - the can and will create zombie processes.

ifireball (changes requested):

splunk/tests/splunk_test_prep.go
Omeramsc: Monitoring the port isn't enough, it's the same issue we had with the shell script in the first stage of the dockerfile... the port & the availability endpoints are up before the service is completely up and ready

ifireball: Still, we should find a network - based approach. The log would not be available when running in a cluster. Using it locally can create all kinds of complications too.

Omeramsc: I implemented the approach from the first comment for now, while I search for a network-based approach. 
but I'm not sure what can we use that we didn't already try in the dockerfile. 
maybe we can loop for 2 min trying query splunk for `eventcount index=test_index` and see that we have *something*  in, but it feels kinda backwards

splunk/tests/splunk_test_prep.go
Omeramsc: If we have the container ID, maybe we can stop it during teardown, using TestMain https://medium.com/goingogo/why-use-testmain-for-testing-in-go-dafb52b406bc ?

splunk/tests/splunk_test_prep.go
ifireball: how do we know the service is not ready when the availability endpoint says it does?

splunk/tests/splunk_test_prep.go
ifireball: the problem with having a fixed container ID si that the user may already have something there. We need to stay out of the user's way as much as possible.

splunk/tests/splunk_test_prep.go
Omeramsc: I tested it before, where a script checked for the availability endpoint (which responded positively before the playbook finished), and proceeded to do data-indexing only to get an error.

splunk/tests/splunk_test_prep.go
Omeramsc: Not sure I get it. how's the container ID fixed if we can get it when first creating it?

splunk/tests/splunk_test_prep.go
ifireball: well perhaps it will behave differently now that we're not indexing data but reading it

splunk/tests/splunk_test_prep.go
ifireball: I was thinking you mean a container name. In any case, for any resource that is in a limited, shared namespace such as port numbers, image names, container names, etc. We must make sure we never collide with something the user may already have there, including another parallel run of our own code.

As far as cleanup goes - I thought about this and realize we' can probably have a single test with multiple subtests, so something along the lines of what I've made for Segment where we have a function that defers the cleanup and calls the test code in a function passed to it may work just as well for this case.

splunk/tests/splunk_test_prep.go
Omeramsc: It seems like we can use `https://localhost:8089/services/server/status`
while it returns 200 a few steps before the ansible playbook finish running, , running a query at the same time/stage returns the expected result.

splunk/tests/splunk_test_prep.go
Omeramsc: I'll try implementing it, 
for now we'll have to verify the port number at the very least. We are currently getting the Container ID when running the container, so at least that's dealt with

splunk/tests/splunk_test_prep.go
Omeramsc: I read a bit, not sure how kube play is coming in hand here? 
also can't we assume it when we build it a moment before? are we trying to avoid a race condition?

ifireball: kube play can automatically build a missing image when an image reference is a local path

> Kube play is capable of building images on the fly given the correct directory layout and Containerfiles. This option is not available for remote clients, including Mac and Windows (excluding WSL2) machines, yet. Consider the following excerpt from a YAML file:

splunk/tests/splunk_test_prep.go
ifireball: I thought that would be the case, the availability signal belongs to the querying front-end not the back-end infestation engine

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,65 @@
...
+func VerifyServiceIsUp() {
+	timeoutStart := time.Now().Unix()
+	splunkURL := "https://localhost:8089/services/server/status"
+	username := "admin"
+	password := "Password"
+
+	for {
+		cmd := exec.Command("curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", "-u",
+			fmt.Sprintf("%s:%s", username, password), "--insecure", splunkURL)
+		output, _ := cmd.CombinedOutput()
```
srivickynesh, line 48:
```suggestion
		output, err := cmd.CombinedOutput()
```
we need to handle this error.

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,65 @@
...
+	return containerID
+}
+
+// VerifyServiceIsUp continuously monitors the Splunk container to
+// ensure it is up and ready for use in tests. It achieves this by repeatedly
+// checking the status endpoint of the Splunk API over a period of two minutes.
+func VerifyServiceIsUp() {
+	timeoutStart := time.Now().Unix()
+	splunkURL := "https://localhost:8089/services/server/status"
+	username := "admin"
```
srivickynesh, line 42: Sensitive Information such as username and password should not be used directly in our code.

Omeramsc: Still a draft, but Is it sensetive? those are the default credentials for the Splunk instance we're getting up just for the tests. you can find those details in `pre-run.sh` and `log_indexing.sh` as well

splunk/tests/splunk_test_prep.go
Omeramsc: Do we? It'll return an error as long as it's not getting 200. meaning the log will be filled with exit codes over and over until the loop will resolve either after 2 minutes or by getting 200.

splunk/tests/splunk_test_prep.go
srivickynesh: Then could you move these to const?

Omeramsc: Yes, Will be moved eventually

splunk/tests/splunk_test_prep.go
Omeramsc: Instead of setting image or container names, I'm having podman setting them up randomly so we won't have any issues with those shared namespaces. 
But checking the port number introduces a race condition between the check and running the container...

splunk/tests/splunk_test_prep.go
Omeramsc: I added the rm flag, do you recommend getting the `podman run` process ID and checking if it completed?

splunk/tests/splunk_test_prep.go
Omeramsc: done

splunk/tests/splunk_test_prep.go
Omeramsc: Seems like things work a bit different using kube play, there the second stage of the dockerfile must kill the splunkd process from the first stage, otherwise the 8089 port stays occupied

Omeramsc: Changed to use podman kube play. This demanded killing the splunkd process during the second stage of the build, as mentioned in the former comment. 
only issue left with this approach is that the manifest has to have a name for the pod - meaning we won't be able to use it concurrently.

splunk/tests/splunk_test_prep.go
```diff
@@ -0,0 +1,129 @@
...
+		"webPort": webPort,
+		"apiPort": apiPort,
+	}
+}
+
+// Building and Running a playbook using 'podman kube play' command.
+// It will make up to three attempts to run with different configurations,
+// allowing it to run concurrently if necessary
+func BuildAndRunPod() map[string]string {
+	yamlContent, err := ioutil.ReadFile("playbook.yaml")
```
srivickynesh, line 49:
```suggestion
	yamlContent, err := os.ReadFile("playbook.yaml")
```
ioutil.ReadFile is deprecated.

Omeramsc: Thanks, updated

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,121 @@
...
+		output, err := cmd.CombinedOutput()
+		if err == nil {
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
+			deployment["containerID"] = containerID
+			return deployment
+		}
+		if retry < retries-1 {
+			log.Printf("Failed to start pod. Retrying (attempt %d of %d)...\n", retry+2, retries)
+		}
```
yftacherzog, line 68: You can make the code simpler by eliminating this if-statement.
```suggestion
		log.Printf("Failed to start pod (attempt %d of %d).\n", retry+1, retries)
```

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+		cmd := exec.Command("podman", "play", "kube", "-")
+		cmd.Stdin = strings.NewReader(deployment["yaml"])
+		output, err := cmd.CombinedOutput()
+		if err == nil {
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
+			deployment["containerID"] = containerID
+			return deployment
+		}
+		log.Printf("Failed to start pod. Retrying (attempt %d of %d)...\n", retry+1, retries)
```
yftacherzog, line 66: My suggesting was to have the attempt number reflect the number of the current failed attempt, and this is why I suggested doing a +1 instead of a +2 while droping the word `Retrying`.
```suggestion
		log.Printf("Failed to start pod (attempt %d of %d).\n", retry+1, retries)
```

Omeramsc: I was pretty sure I removed it, git says otherwise... will fix

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+var manifest string
+
+const Username = "admin"
+const Password = "Password"
+
+// Generates a random dynamic port number
+func GetRandomPort() string {
+	minPort := 49152
+	maxPort := 65535
+	rand.Seed(time.Now().UnixNano())
```
srivickynesh, line 23: rand.Seed is deprecated

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
+			deployment["containerID"] = containerID
+			return deployment
+		}
+		log.Printf("Failed to start pod (attempt %d of %d).\n", retry+1, retries)
+	}
+	Cleanup("", deployment["podName"])
+	log.Fatalf("Failed to start pod after %d retries", retries)
+	return make(map[string]string) // Avoid 'missing return' error
```
amisstea, line 70: Why not return an error?

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+		log.Println("Stopping and removing the container...")
+		cmd := exec.Command("podman", "stop", containerID)
+		if err := cmd.Run(); err != nil {
+			log.Println("Error stopping the container:", err)
+		}
+		cmd = exec.Command("podman", "rm", "-f", containerID)
+		if err := cmd.Run(); err != nil {
+			log.Println("Error removing the container:", err)
+		}
+	}
```
amisstea, line 107: Why do we need this if `podName` is known and included in the return value of `BuildAndRunPod`?

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+	for retry := 0; retry < retries; retry++ {
+		deployment := modifyYaml(manifest)
+
+		cmd := exec.Command("podman", "play", "kube", "-")
+		cmd.Stdin = strings.NewReader(deployment["yaml"])
+		output, err := cmd.CombinedOutput()
+		if err == nil {
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
+			deployment["containerID"] = containerID
```
amisstea, line 63: Why do we need to get the `containerID`?

containerfixture/containerfixture.go
Omeramsc: Only to shut it gracefully, I wanted to avoid removing them forcefully with `podman pod rm -f`. We can remove that, not sure if we shuld.

containerfixture/containerfixture.go
Omeramsc: Currently  only for the cleanup

containerfixture/containerfixture.go
Omeramsc: `log.Fatalf` already  call to os.Exit(1), the `return.make` will never be reached (but GO requires it). how would you suggest to change it?

containerfixture/containerfixture.go
Omeramsc: fixed

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,119 @@
...
+		"apiPort": apiPort,
+	}
+}
+
+// Building and Running a k8s manifest using 'podman kube play' command.
+// It will make up to three attempts to run with different configurations,
+// allowing it to run concurrently if necessary
+func BuildAndRunPod() map[string]string {
+	var deployment map[string]string
+	retries := 3
```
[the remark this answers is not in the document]
Omeramsc: fixed

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+	maxPort := 65535
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	randomPort := fmt.Sprintf("%d", r.Intn(maxPort-minPort)+minPort)
+	return randomPort
+}
+
+// Modify the manifest to allow running it in parallel as
+// many times as needed with dynamic values
+func modifyYaml(yaml string) map[string]string {
```
ifireball, line 31: GO has real structs, no reason to use maps in their place.

You could also just return multiple outputs.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+	"log"
+	"math/rand"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+//go:embed containerfixture.yaml
+var manifest string
+
```
ifireball, line 15: Do these consts need to be exported? If not - start with lowercase letters.

containerfixture/containerfixture.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  containers:
+  - args:
+    - start-service
+    env:
+    - name: SPLUNK_START_ARGS
+      value: --accept-license
+    - name: SPLUNK_PASSWORD
+      value: Password
+    image: splunk
```
ifireball, line 15: the file name suggests this is meant to start any container - but the image is fixed to the splunk image...

The ports section may also need to be different for different containers.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+const Username = "admin"
+const Password = "Password"
+
+// Generates a random dynamic port number
+func GetRandomPort() string {
```
ifireball, line 20: Does this function need to be exported?

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+	return randomPort
+}
+
+// Modify the manifest to allow running it in parallel as
+// many times as needed with dynamic values
+func modifyYaml(yaml string) map[string]string {
+	podName := fmt.Sprintf("test-pod-%s", GetRandomPort())
+	webPort := GetRandomPort()
+	apiPort := GetRandomPort()
+
```
ifireball, line 35: perhaps we could use go templates here instead of simply replacing fixed strings - we could add a custom template function for getting a random port number while also storing it in a backing map.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+	attempts := 3
+	for attempt := 0; attempt < attempts; attempt++ {
+		deployment := modifyYaml(manifest)
+
+		cmd := exec.Command("podman", "play", "kube", "-")
+		cmd.Stdin = strings.NewReader(deployment["yaml"])
+		output, err := cmd.CombinedOutput()
+		if err == nil {
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
```
ifireball, line 62: It dangerous to assume the container ID would necessarily be the last line of output - especially since you combined STDOUT and STDERR.

It may be better to have STDERR go to the test logs.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// VerifySplunkServiceIsUp continuously monitors the Splunk container to
+// ensure it is up and ready for use in tests. It achieves this by repeatedly
+// checking the status endpoint of the Splunk API over a period of two minutes.
+func VerifySplunkServiceIsUp(port string) {
+	timeoutStart := time.Now().Unix()
+	splunkURL := fmt.Sprintf("https://localhost:%s/services/server/status", port)
+	for {
+		cmd := exec.Command("curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", "-u",
```
ifireball, line 80: It would be simpler to use `net/http` for this

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	Cleanup("", deployment["podName"])
+	log.Fatalf("Failed to start pod after %d attempts", attempts)
+	return make(map[string]string) // Avoid 'missing return' error
+}
+
+// VerifySplunkServiceIsUp continuously monitors the Splunk container to
+// ensure it is up and ready for use in tests. It achieves this by repeatedly
+// checking the status endpoint of the Splunk API over a period of two minutes.
+func VerifySplunkServiceIsUp(port string) {
```
ifireball, line 76: This could have a generalized name sine there is nothing Splunk-specific here.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+		cmd := exec.Command("podman", "play", "kube", "-")
+		cmd.Stdin = strings.NewReader(deployment["yaml"])
+		output, err := cmd.CombinedOutput()
+		if err == nil {
+			outputString := strings.Split(strings.TrimSpace(string(output)), "\n")
+			containerID := outputString[len(outputString)-1]
+			deployment["containerID"] = containerID
+			return deployment
+		}
+		log.Printf("Failed to start pod (attempt %d of %d).\n", attempt+1, attempts)
```
ifireball, line 66: This is meant to be a test fixture - pass in a testing.T structure and use its logging functions instead of a generic logger.  Or even better - use assertions.

If you want the code to be generic - define an interface with the logging methods you need so that it can fit both `testing.T` and `logging/Logger` (It seems most or their methods have identical signatures anyway).

containerfixture/containerfixture.go
ifireball: This should be failing the test not the whole testing process.

splunk/Dockerfile
```diff
@@ -26,7 +26,7 @@ EXPOSE 8089
 
 USER root
 
-RUN mkdir -p /opt/splunk/var/lib/splunk
+RUN mkdir -p /opt/splunk/var/lib/splunk && pkill -f "splunkd"
```
ifireball, line 29: no way Splunk is running here if you didn't start it - I guess you miss -diagnosed some other problem.

Are you properly stopping Splunk in that pre-run script?

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+		if time.Now().Unix()-timeoutStart > 120 {
+			log.Fatalf("The Splunk instance is not up, cannot verify indexing for tests.")
+		}
+		time.Sleep(5 * time.Second)
+	}
+}
+
+// Making sure to stop and remove the container we deployed
+func Cleanup(containerID string, podName string) {
+	if containerID != "" {
```
ifireball, line 96: Why do we need to remove the container on its own if we're going to remove the entire pod?

ifireball (changes requested):
This code is not mature enough to be merged. 

One notable missing thing is a wrapper function to bring the fixture up, run a test and then shut it down.

containerfixture/containerfixture.go
Omeramsc: The endpoint itself is Splunk-specifc, as well as most of the cmd used (`fmt.Sprintf("%s:%s", Username, Password), "--insecure", splunkURL)`). modifying this alongside the logging make the function too messy iMO

splunk/Dockerfile
Omeramsc: The Process was running in the background. 
We already went over the fact that stopping Splunk manually during the first stage of the build, for some reason, fails the whole deployment at the end. reason unknown. that's why we've gone with the multistage in the first place. 
If I'd have to guess, the port probably stays occupied, but I can't remember if it was confirmed.

containerfixture/containerfixture.go
Omeramsc: Only to have a graceful removal. if you think it's redundant, I don't mind removing it. It'll make things cleaner. Do note removing the pod sometimes does not resolve the port forwarding, if we care.

containerfixture/containerfixture.go
Omeramsc: I mostly wanted to avoid using regex or something similar...

containerfixture/containerfixture.go
Omeramsc: I'll check it out

containerfixture/containerfixture.go
Omeramsc: Currently not, but a Splunk test will need the Username and Password detaills, unless those are sent to the function.
I'd say the other way around is more suitable (them being imported FROM the Splunk package), but that'll create circular dependencies.

splunk/Dockerfile
ifireball: The fact you end up having a process up means there is a bug in the build process that its not starting new containers for different build stages. 

We've gone with multistage so that any changes we don't want that are made in the build image do not make it to the final image.

The only thing that could mess things up for us here is if killing Splunk causes it to delete or corrupt the files we are copying to the final image. If that is the case, we can have a separate "clean" stage to kill the process. In any case this does not belong in a layer in the final image.

containerfixture/containerfixture.go
ifireball: So potentially have garbage instead? In any case perhaps we can drop getting the ID here entirely.

containerfixture/containerfixture.go
ifireball: that Sprintf shouldn't be needed - we should have a netrc file taking care of that - easier and more reliable to have an env var pointing to a static file , and we would need the same for the test anyway.

The endpoint URL could be passed as a parameter.

containerfixture/containerfixture.go
ifireball: Unless you use netrc as suggested above.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,118 @@
...
+
+// VerifySplunkServiceIsUp continuously monitors the Splunk container to
+// ensure it is up and ready for use in tests. It achieves this by repeatedly
+// checking the status endpoint of the Splunk API over a period of two minutes.
+func VerifySplunkServiceIsUp(port string) {
+	timeoutStart := time.Now().Unix()
+	splunkURL := fmt.Sprintf("https://localhost:%s/services/server/status", port)
+	for {
+		cmd := exec.Command("curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", "-u",
+			fmt.Sprintf("%s:%s", Username, Password), "--insecure", splunkURL)
```
ifireball, line 81: The tested code would not be using `--insecure` - we need to solve this - possibly by hard injecting our own certificates into the Splunk image so it uses those and setting `curl` to use them as well while testing.

Alternatively, could we talk to Splunk over HTTP?

containerfixture/containerfixture.go
ifireball: pass them as parameters, or otherwise us a netrc  file as suggested below.

In general if we can use files or env vars to avoid having the code deal with credentials its almost always better.

containerfixture/containerfixture.go
ifireball: Removing a POD is not IMO less graceful then removing the containers in it first. Also we're using a POD to make Podman do more things for us instead of us having to write code to do them - so its exactly the kind of code this is meant to let us avoid writing.

containerfixture/containerfixture.go
ifireball: How is a generic fixture package depending on a Specific "Splunk" package that is using it "more suitable"?

containerfixture/containerfixture.go
srivickynesh: type DeploymentInfo struct {
	Yaml        string
	PodName     string
	WebPort     string
	ApiPort     string
}
```suggestion
func modifyYaml(yaml string) (*DeploymentInfo, error) {
```

containerfixture/containerfixture.go
Omeramsc: It does not depend on a Splunk package, but if the function will get generlized, and will get used on the Splunk test, it might just get the`username` and `password` as parameters, without importing them.   
anyway, I already said below I'll check the netrc/parameters approach

splunk/Dockerfile
Omeramsc: This issue raise only when using `kube play kube`. my guess is both stages share the same network namespace, which is why the ports from the first stage stay occupied and the process linger. 
I did not find a different solution for it nor did I find a method to close Splunk using commands. 
I thought `RUN /bin/bash /opt/splunk/bin/splunk stop` will do the trick but it only works when I actually enter a Splunk container, otherwise it fails. 

I might resort to another layer just to kill the service, although I'd prefer not to...

containerfixture/containerfixture.yaml
Omeramsc: Not sure how can we generalize this file. it's not only the image name, but the args, ports, and env vars. we might as well just have a different yaml for every service (or have one for all of them, but containing multiple definitions, not dynamic ones)

ifireball: Using a template can help here - each fixture can have its own template

Omeramsc: I've implemented templates into it, but I don't see how replacing complete sections is readable or better than having separate files...

containerfixture/containerfixture.go
Omeramsc: fixed

containerfixture/containerfixture.go
Omeramsc: fixed, replaced with `t.Fail()` alongside `t.Logf`

containerfixture/containerfixture.yaml
ifireball: I did say those were mutually exclusive

containerfixture/containerfixture.go
Omeramsc: using a .netrc file

containerfixture/containerfixture.go
Omeramsc: settled with removing the pod with -f.

containerfixture/containerfixture.go
Omeramsc: using testing.T for now.

containerfixture/containerfixture.go
Omeramsc: Applied simple templating, will see to the backing map

containerfixture/containerfixture.yaml
Omeramsc: I can implement a separate template for every fixture; but since we cannot embed dynamically, it's either embedding each file separately (and then somehow using the specific one we need?) or embedding a filesystem, which I'm not sure is desirable

splunk/Dockerfile
Omeramsc: Solved! successfully closed the server from within pre-run

containerfixture/containerfixture.go
Omeramsc: splunkd does not communicate in HTTP, so no. 
I'll have to figure out the first part of what you suggested.

containerfixture/containerfixture.yaml
ifireball: Embed to the image-specific fixture and pass as a parameter to the generic "base" fixture.

Omeramsc:
Improved the code a bit and added a simple Splunk test. 
handling certificates or a good way to handle the path to .netrc have not been implemented yet.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,131 @@
...
+			t.Logf("STDOUT: %s", stdoutBuf.String())
+			return deployment
+		}
+		t.Logf("Failed to start pod (attempt %d of %d).\n", attempt+1, attempts)
+		t.Logf("STDERR: %s", stderrBuf.String())
+	}
+	Cleanup(t, deployment["podName"])
+	t.Fail()
+	t.Logf("Failed to start pod after %d attempts", attempts)
+	return DeploymentInfo{"", "", "", ""} // Avoid 'missing return' error
```
ifireball, line 81: No need to manually specify the empty struct fields - the zero value of a string is "" anyway.

splunk/splunk_test.go
```diff
@@ -0,0 +1,64 @@
...
+		host       string
+		port       string
+		index      string
+		want       string
+		want_error bool
+	}{
+		{
+			name:  "Simple query",
+			host:  "localhost",
+			port:  deployment.ApiPort,
```
ifireball, line 41: If the port is always `deployment.ApiPort` why include that in the test struct?

splunk/splunk_test.go
```diff
@@ -0,0 +1,64 @@
...
+			index: "test_index",
+			want:  "10",
+		},
+		{
+			name:       "Simple query with non-existing host",
+			host:       "fake-host",
```
ifireball, line 47: I don't see how is testing that you cannot query the wrong host useful for testing the fixture for bringing up the host... This is testing `GetRecordsCount`...

ifireball (changes requested):
I think there is a need for a couple of top-level fixture function - one generic one that:

1. brings up a pod
2. waits for the service to start
3. runs a test function passed as a parameter
4. cleans up reliably

And another Splunk-specific one that does this just for Splunk (using the generic one).

splunk/splunk_test.go
Omeramsc: removed for now, but I'm trying to address the errors in the code. since the host is a parameter, giving a wrong value for one of the parameters (anything but index or outputMode) will result in a failure. why not check that it works as expected?

splunk/splunk_test.go
```diff
@@ -0,0 +1,64 @@
...
+//go:embed splunk_container_template.yaml
+var manifest string
+
+func TestGetRecordsCount(t *testing.T) {
+	_, callerFilePath, _, _ := runtime.Caller(0)
+	callerDirPath := filepath.Dir(callerFilePath)
+	netrcPath := fmt.Sprintf("%s/.netrc", callerDirPath)
+	fmt.Println(netrcPath)
+	
+
```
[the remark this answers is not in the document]
Omeramsc: Achievable, but unless we run everything through this fixture function (is this what we want?) , we'll need to defer the cleanUp function outside of it, since we need the service open until after this function is done running.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,131 @@
...
+// VerifyServiceIsUp continuously monitors the service container to
+// ensure it is up and ready for use in tests. It achieves this by repeatedly
+// checking the status endpoint of the service API over a period of two minutes.
+func VerifyServiceIsUp(t *testing.T, endpoint string, serviceName string) {
+	netrcPath, _ := GetNetrcPath()
+	timeoutStart := time.Now().Unix()
+	for {
+		cmd := exec.Command("curl", "-s", "-o", "/dev/null", "-w", "%{http_code}",
+			"--netrc-file", netrcPath, "--insecure", endpoint)
+		output, _ := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
Omeramsc: fixed, made it error when there's an error that is *not* is http code "000"  (as this can repeat until the service is up)

splunk/splunk_test.go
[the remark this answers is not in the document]
Omeramsc: disregard my last comment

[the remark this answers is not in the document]
Omeramsc: I'm not completely sure how to assemble this in a way that'll run the fixture only once during a parametrize test (is there a reason to deploy multiple pods just for another testcase in the same test?)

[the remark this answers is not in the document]
ifireball: Think of this like a Python context manager - in Python when you create a resource that you need to clean up, you make a context manager to manage that and you wrap everything that needs the resource in it.

In Go you achieve pretty much the same via a function that defers the cleanup an calls another closure/function that is passed to it.

WRT the test parameter loops - you can either wrap the entire loop with the "context manager" or wrap each individual subtest with it, depending on wither you want to clean the resources between subtests.

splunk/splunk_test.go
ifireball: that would be a failure of the code talking to the fixture, not a failure of the fixture itself.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,133 @@
...
+func getRandomPort() string {
+	minPort := 49152
+	maxPort := 65535
+	randomPort := fmt.Sprintf("%d", r.Intn(maxPort-minPort)+minPort)
+	return randomPort
+}
+
+// Modify the yaml manifest to allow running it in parallel
+// as many times as needed with dynamic values for its name and ports
+func ApplyDynamicDeploymentTemplate(yaml string) DeploymentInfo {
```
ifireball, line 38: if this is a "dynamic deployment" template, what would be a "static" one? also is this a template of a deployment? deployment to where? a production cluster? Lets try to be specific about what we are doing:

```suggestion
func GenerateFixturePodYaml(template string) FixtureInfo {
```

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,133 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"text/template"
+	"time"
+)
+
+const waitTimeoutSeconds = int64(120)
```
ifireball, line 16: "wait timeout" is both redundant and opaque - try saying what the timeout is for.

Also you don't need to cast the constant - Go will do it for you when you use it.

splunk/splunk.go
```diff
@@ -0,0 +1,14 @@
+package splunk
+
+import (
+	"fmt"
+)
+
+// CalculateAPIEndpoint builds a valid endpoint to be used against a
+// plunk Search API service
+func CalculateSearchAPIEndpoint(host, port, owner, app, outputMode string) string {
```
ifireball, line 9:
```suggestion
func GetSearchAPIEndpoint(host, port, owner, app, outputMode string) string {
```

They way this is defined it would not be useful to get the value to pass into `SPLUNK_APP_API_URL` please separate between getting the general Splunk API URL, and getting the URL for the specific search endpoint.

Also I don't think `outputMode` should be hardwired into the URL like this, its cleaner to pass it via `-d` anyway.

ifireball (changes requested):

splunk/splunk.go
Omeramsc: changed and added functions to match `fetch-uj-records`

containerfixture/containerfixture.go
Omeramsc: fixed

splunk/splunk_test.go
```diff
@@ -0,0 +1,76 @@
...
+	cmd := exec.Command("curl", "--netrc-file", netrcPath, SplunkAppSearchURL, "-d", searchQuery, "-d", "output_mode=csv")
+
+	output, err := cmd.CombinedOutput()
+
+	if err != nil {
+		log.Printf("Failed to count %s index records: %v", index, err)
+		return "", err
+	}
+
+	re := regexp.MustCompile(`(?m)^count\s+(\d+)$`)
```
ifireball, line 41: Might as well compile this and crush on module load and not on function call.

splunk/splunk_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func GetRecordsCount(host, port, owner, app, index string) (string, error) {
+	splunkAppApiURL := GetSplunkAppAPIEndpoint(host, port, owner, app)
+	SplunkAppSearchURL := GetSearchAPIEndpoint(splunkAppApiURL)
+	searchQuery := "search=" + fmt.Sprintf("search index=%s | stats count", index)
+	netrcPath, _ := containerfixture.GetNetrcPath()
+	cmd := exec.Command("curl", "--netrc-file", netrcPath, SplunkAppSearchURL, "-d", searchQuery, "-d", "output_mode=csv")
+
+	output, err := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
Omeramsc: fixed.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,143 @@
...
+
+type FixtureInfo struct {
+	Yaml    string
+	PodName string
+	WebPort string
+	ApiPort string
+}
+
+// Generates a random dynamic port number
+func getRandomPort() string {
```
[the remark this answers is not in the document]
Omeramsc: Your refer to minPort and maxPort? this is the range for private/dynamic ports, it is not relevant to the package, but to this function specifically, as it creates a dynamic ports.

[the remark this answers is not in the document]
ifireball: its customary to define consts at the top of the file - this function is a part of the overall package so the consts are relevant for the package.

In any case consts should be consts and not variable unless there is a reason they shouldn't be.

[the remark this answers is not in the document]
Omeramsc: set as consts

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := cmd.Run(); err != nil {
+		t.Errorf("Error stopping the pod: %s", err)
+	}
+	removePodCmd := exec.Command("podman", "pod", "rm", "-f", podName)
+	if err := removePodCmd.Run(); err != nil {
+		t.Fatalf("Error removing the pod: %s", err)
+	}
+}
+
+func WithServiceContainer(t *testing.T, manifest string, test_func func(*testing.T, FixtureInfo), msgAndArgs ...interface{}) {
```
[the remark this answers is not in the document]
ifireball: the functionality of bringing a container up and down seems to be quite unrelated to the "up" check which needs to know details about the service API. The only relates detail between the two is the port number.

I would decouple them - and bring them together in the service-specific fixtures. That would allow for more flexibility in the long term such as having containers with multiple services.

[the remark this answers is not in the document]
Omeramsc: I get why you'd want to decouple the pod itself from the service (although it is specifically a service container fixture). 
The way I see for this decoupling is making `WithServiceContainer` a very simple fixture to create the pod and get the deployment data, 
while `WithSplunkContainer` run the former and does all the heavy lifting (service check, cleanup, run tests). 
Did you have anything else in mind?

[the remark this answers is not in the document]
ifireball: since the service details such as what the api looks like, how many port etc. is a service-specific thing - we can have the service  fixture put the checking and the bringing up/down together while keeping those as separate functions in the generic fixture module.

splunk/splunk_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+//go:embed splunk_container_template.yaml
+var splunkServiceManifest string
+var countPattern = regexp.MustCompile(`(?m)^count\s+(\d+)$`)
+
+const NotUpErrorMsg = "The %s instance is not up, cannot verify indexing for tests."
+
+func WithSplunkContainer(t *testing.T, test_func func(*testing.T, containerfixture.FixtureInfo)) {
+	containerfixture.WithServiceContainer(t, splunkServiceManifest, test_func)
```
ifireball, line 24:
```suggestion
containerfixture.WithServiceContainer(
    t, splunkServiceManifest, func(t *testing.T, fi containerfixture.FixtureInfo) {
        endpoint := fmt.Sprintf("http://localhost:%s/%s", fi.ApiPort, ServiceStatusCheckPath)
        containerfixture.RequireServiceIsUp(t, endpoint, ServiceName, NotUpErrorMsg)
        test_func(t, fi)
    })
```

Also this kind of makes you see that you don't really need to be explicitly passing 't' into the test func in both fixtures because when defined as a closure it can always get `t` from the surrounding function that defines it.

Also small styling nitpicks - in Go you use `camelCase` and not snake_case` for variables. And you don't seem to be using consistent indentation, please make sure you use 4 spaces.

containerfixture/containerfixture.go
```diff
@@ -0,0 +1,138 @@
...
+	"strings"
+	"testing"
+	"text/template"
+	"time"
+)
+
+const serviceUpTimeout = 120
+const containerBuildAttempts = 3
+const minDynamicPort = 49152
+const maxDynamicPort = 65535
```
srivickynesh, line 19: const (
	serviceUpTimeout        = 120
	containerBuildAttempts  = 3
	minDynamicPort          = 49152
	maxDynamicPort          = 65535
)

srivickynesh (changes requested):

splunk/splunk_test.go
Omeramsc: Fixed, thanks for the explanation, it is more clear now! 
Obviously as seen by the rest of the code, I'm aware of the camelCase... but old habits die hard. will try to notice it more.

containerfixture/containerfixture.go
Omeramsc: fixed

ifireball (changes requested):
As mentioned multiple time before - please fix you indentation - run gofmt on the code if you are not doing that already and make sure you configure your editor to use 4 spaces and not tabs or whatever it is using ATM.

Omeramsc:
> As mentioned multiple time before - please fix you indentation - run gofmt on the code if you are not doing that already and make sure you configure your editor to use 4 spaces and not tabs or whatever it is using ATM.

I already ran gofmt on all files, and visual studio is configured to 4 spaces. I don't know where the issue lies.

ifireball:
> > As mentioned multiple time before - please fix you indentation - run gofmt on the code if you are not doing that already and make sure you configure your editor to use 4 spaces and not tabs or whatever it is using ATM.
> 
> I already ran gofmt on all files, and visual studio is configured to 4 spaces. I don't know where the issue lies.
> 
> code in my editor:
> 
> ```
> type FixtureInfo struct {
>     Yaml    string
>     PodName string
>     WebPort string
>     ApiPort string
> }
> ```
> 
> code here, for some reason:
> 
> ```
> type FixtureInfo struct {
> 	Yaml    string
> 	PodName string
> 	WebPort string
> 	ApiPort string
> }
> ```

One thing that can cause this is if:

1. You are using the tab key in your editor or have otherwise configured it to indent with tabs
2. Your editor is configured to present tabs as 4-space wide
3. Your editor is not configured to covert tabs to spaces.

Omeramsc:
> > > As mentioned multiple time before - please fix you indentation - run gofmt on the code if you are not doing that already and make sure you configure your editor to use 4 spaces and not tabs or whatever it is using ATM.
> > 
> > 
> > I already ran gofmt on all files, and visual studio is configured to 4 spaces. I don't know where the issue lies.
> > code in my editor:
> > ```
> > type FixtureInfo struct {
> >     Yaml    string
> >     PodName string
> >     WebPort string
> >     ApiPort string
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > code here, for some reason:
> > ```
> > type FixtureInfo struct {
> > 	Yaml    string
> > 	PodName string
> > 	WebPort string
> > 	ApiPort string
> > }
> > ```
> 
> One thing that can cause this is if:
> 
> 1. You are using the tab key in your editor or have otherwise configured it to indent with tabs
> 2. Your editor is configured to present tabs as 4-space wide
> 3. Your editor is not configured to covert tabs to spaces.

All seems to be configured correctly; also, I ran go fmt (after closing the IDE) which should emit the tabs anyway...

ifireball:
> 
> All seems to be configured correctly; also, I ran gofmt (after closing the IDE) which should emit the tabs anyway...

Not sure wither gofmt removes tabs or not - you can use some HEX editing software to see exactly what you have in the file. The files here do appear differently **in GitHub** then other files in the repo so I don't think we can blame GitHub for miss-displaying things.

ifireball:
> > All seems to be configured correctly; also, I ran gofmt (after closing the IDE) which should emit the tabs anyway...
> 
> Not sure wither gofmt removes tabs or not - you can use some HEX editing software to see exactly what you have in the file. The files here do appear differently **in GitHub** then other files in the repo so I don't think we can blame GitHub for miss-displaying things.

Ok, nvm it seems that gofmt forces tabs - and we have them in all the files in the repo, GitHub is just not consistant about displaying them it seems.

ifireball (approved):
Approving since IIRC we discussed everything already

srivickynesh (approved):
LGTM

hmariset (approved):

PR #16, opened 2023-07-16
feat(RHTAPWATCH-374): Run multiple Splunk queries

ifireball:
Add support in `querygen` for generating and printing multiple queries as well as support in `fetch-uj-records.sh` run running them.

In the process also add one additional query for querying PipelineRuns

PR #21, opened 2023-07-20
feat(RHTAPWATCH-372): Map resource namespaces to SSO UIDs

amisstea:
Some resources don't have a userId field. This change introduces a fallback mechanism to lookup the SSO UID for the workspace owner in such cases.

### DIFF
--- a/scripts/splunk-to-segment.sh
+++ b/scripts/splunk-to-segment.sh
@@ -85,17 +85,25 @@ function event_subject_map() {
   }'
 }
 
+# Not all event resources have a userId field necessary for attribution in Segment.
+# In such cases, the owner of the workspace is used instead.
+# For this to work workspaces must be named after a valid SSO username.
 jq \
   --compact-output \
   --slurpfile uidm <(get-uid-map.sh) \
+  --slurpfile wksm <(get-workspace-map.sh) \
   --slurpfile evvm <(event_verb_map) \
   --slurpfile evsm <(event_subject_map) \
-  '.result | select($uidm[0][.userId])
+  '.result
+  | .userIdOrWorkspace = (.userId // $wksm[0][.namespace])
+  | select(.userIdOrWorkspace)
+  | .ssoId = $uidm[0][.userIdOrWorkspace]
+  | select(.ssoId)
   | {
       messageId, 
       timestamp, 
       type, 
-      userId: $uidm[0][.userId], 
+      userId: .ssoId,
       event: "\($evsm[0][.event_subject] // .event_subject) \($evvm[0][.event_verb])", 
       properties: (.properties|fromjson)
     }


amisstea:
Marking as a draft until https://github.com/redhat-appstudio/segment-bridge/pull/17 is merged.

querygen/querygen.go
```diff
@@ -20,7 +20,7 @@ const (
 	// Segment record is quite fixed, most queries should return the same
 	// fields.
 	IncludeFieldsCmd = `fields ` +
-		`messageId, timestamp, type, userId, event_verb, event_subject, properties`
+		`messageId, timestamp, type, userId, event_verb, event_subject, properties, namespace`
```
ifireball, line 23: This was already done in PR #16

querygen/querygen.go
```diff
@@ -73,6 +73,7 @@ func GenTrackFields(spec TrackFieldSpec, properties_map map[string]string) strin
 	builder.WriteString(`messageId=auditID,`)
 	builder.WriteString(`timestamp=requestReceivedTimestamp,`)
 	builder.WriteString(`type="track",`)
+	builder.WriteString(`namespace='objectRef.namespace',`)
```
ifireball, line 76: This was already done in PR #16

scripts/splunk-to-segment.sh
```diff
@@ -85,17 +85,23 @@ function event_subject_map() {
...
 jq \
   --compact-output \
   --slurpfile uidm <(get-uid-map.sh) \
+  --slurpfile wksm <(get-workspace-map.sh) \
   --slurpfile evvm <(event_verb_map) \
   --slurpfile evsm <(event_subject_map) \
-  '.result | select($uidm[0][.userId])
+  '.result
+  | .userIdOrWorkspace = (.userId // $wksm[0][.namespace])
+  | select(.userIdOrWorkspace and $uidm[0][.userIdOrWorkspace])
```
ifireball, line 99: Instead of doing the lookup in `$uidm` twice, might as well do that once here, place in a variable and then select on that variable.

ifireball (changes requested):
Please remove the parts that were already implemented elsewhere, otherwise lgtm.

querygen/querygen.go
amisstea: Rebased.

querygen/querygen.go
amisstea: Rebased.

scripts/splunk-to-segment.sh
amisstea: Done.

ifireball:
@Omeramsc , @yftacherzog please review, let get this merged.

yftacherzog (approved):

PR #22, opened 2023-07-22
chore(RHTAPWATCH-375): Support Go binaries in our job container image

Kousalya1998:
- Build Go binaries in the first stage
- Copy them from builder in second stage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,16 @@
+# First stage: Build the Go binaries
+FROM registry.redhat.io/rhel8/go-toolset:1.19.10-3 AS builder
+WORKDIR /app
+COPY . .
+RUN go build -o /app/build/ ./cmd/...
+
+# Second stage: Create the final container image
 FROM registry.redhat.io/openshift4/ose-tools-rhel8:v4.13.0-202306070816.p0.g05d83ef.assembly.stream
 
 COPY --chown=root:root --chmod=644 data/ca-trust/* /etc/pki/ca-trust/source/anchors
 RUN /usr/bin/update-ca-trust
 COPY --chown=root:root --chmod=755 scripts/* /usr/local/bin
+COPY --chown=root:root --chmod=755 --from=builder /app/build/* /usr/local/bin/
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,8 +1,16 @@
+# First stage: Build the Go binaries
+FROM registry.redhat.io/rhel8/go-toolset:1.19.10-3 AS builder
+WORKDIR /app
+COPY . .
+RUN cd cmd/querygen && go build -o /app/build/querygen
```
ifireball, line 5: Building this way would require changing the dockerfile each time we add a new binary, can we find a way to build the whole package with a  single command and have all the binaries be places in a single directory?

ifireball (commented):
This gets the job done for now, but perhaps we can make it more resilient to future changes.

Dockerfile
amisstea: This should work (untested):
```suggestion
RUN cd cmd && go build -o /app/build ./...
```

srivickynesh: yep this works (tested)

Kousalya1998: > Building this way would require changing the dockerfile each time we add a new binary, can we find a way to build the whole package with a single command and have all the binaries placed in a single directory?

@ifireball 
When you say add a new binary, do you mean the folder structure is expected to grow like below?

```
cmd/
  cmd1/
    main.go
  cmd2/
    main.go
  cmd3/
    main.go
  ...
```
In that case do we need to build binary with the name of folder name (Eg. cmd1, cmd2) and copy to the /usr/local/bin/?

ifireball: yes, that kind of directory structure seems to be a convention for Go projects with multiple binaries. Hopefully it would simplify the task here

Kousalya1998: okay makes sense

Kousalya1998: > This should work (untested):

@amisstea @srivickynesh Ty.

srivickynesh (approved):
LGTM

ifireball (approved):

PR #23, opened 2023-07-24
A helper module for running the scripts from Go

ifireball:
Make it easier to write tests for them.

Also adding the `testify` module as a dependency to make nice assertions.

As a simple demonstration of testify also changed the tests
for `webfixture`.


scripts/runner.go
```diff
@@ -0,0 +1,37 @@
...
+	if !ok {
+		return "", errors.New("failed to find path of scripts via Go file name")
+	}
+	if err := pushToPath(path.Dir(goFile)); err != nil {
+		return "", err
+	}
+	return exec.LookPath(file)
+}
+
+// Push the given directory in from of $PATH unless its already listed there
```
yftacherzog, line 26: I think there's a typo on the comment. I can't understand this sentence.

scripts/runner.go
```diff
@@ -0,0 +1,37 @@
...
+	osPath := os.Getenv("PATH")
+	osPathList := filepath.SplitList(osPath)
+	for _, pathDir := range osPathList {
+		if dir == pathDir {
+			return nil
+		}
+	}
+	newOsPath := fmt.Sprintf("%s%c%s", dir, filepath.ListSeparator, osPath)
+	return os.Setenv("PATH", newOsPath)
+}
```
yftacherzog, line 37: missing newline at end of file

scripts/runner_test.go
```diff
@@ -0,0 +1,45 @@
...
+		file string
+		wantErr bool
+	}{
+		{"no-such-file", true},
+		{"fetch-uj-records.sh", false},
+		{"get-uid-map.sh", false},
+		{"splunk-to-segment.sh", false},
+		{"segment-mass-uploader.sh", false},
+	}
+	for _, tt := range tests {
```
yftacherzog, line 20: Is there a way to make the intention of the test more clear?
I'm reading this and I cannot tell for sure what we're testing.

scripts/runner.go
```diff
@@ -0,0 +1,37 @@
...
+// It works by manipulating $PATH
+func LookPath(file string) (string, error) {
+	_, goFile, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", errors.New("failed to find path of scripts via Go file name")
+	}
+	if err := pushToPath(path.Dir(goFile)); err != nil {
+		return "", err
+	}
+	return exec.LookPath(file)
```
Omeramsc, line 23: So if I got it right, first we manipulate the PATH to include the directory of this very script, then if everything went well we run `LookPath` which now search through a modified path? 
if so, might want to elaborate on this on the comment doc...

scripts/runner_test.go
ifireball: no sure how, is basically iterating over script names, passing them to LookPath and then checking that the returned string:

1. ends with the script name
2. points to an existent, executable, regular file.

The `if`s are quire straight forward, so I'm not sure how to improve that. Permaps I should rename the `file` variable to indicate better its a script name.

scripts/runner.go
ifireball: done

scripts/runner_test.go
yftacherzog: I think those loop and ifs go against the principles you're advocating for about writing simple test cases.

I'm now thinking that what makes the tests less obvious to comprehend is that we're not asserting for what we're after, but instead checking the absence of the complimenting condition. This also makes the tests less usable as a tool for understanding the code itself.

ifireball: I think most of the things you mention are idioms that arise from how the Go `testing` module is built.

The loop is equivalent to the `pytest.parameterize` decorator, but there are no decorators in Go, and in general, more of a tendency to do things explicitly rather then implicitly via e.g decorators. 

I think this loop is pretty much a standard idiom, given that it is essentially auto-generated for me by VScode. We should view the function more like a complete test suite running individual sub tests via `t.Run` rather then as a single test case.

As for the assertions - I guess I could come up with abstractions to make them look more like the assertions you are used to, but ultimately I think we are better off learning how to read code written to the existing standard library API.

I'm willing to try concrete suggestions, if you have any.

yftacherzog (approved):

scripts/runner_test.go
yftacherzog: https://pkg.go.dev/github.com/stretchr/testify/assert

I see this package for assertions which I think might make things look more straightforward. 
I don't think we should hold merging this one at this point, but I think this is something that could make our tests and code more maintainable.

In such case, we can replace this:
```
if !tt.wantErr {
	t.Errorf("LookPath() error = %v, wantErr %v", err, tt.wantErr)
}
```
With this:
```
assert.Truef(tt.wantErr, "LookPath() error = %v, wantErr %v", err, tt.wantErr)
```

There are assertions there for errors that can probably be used in this case which are more appropriate than asserting for a boolean value, but it's already more straightforward as far as I see it.

srivickynesh (approved):
lgtm

scripts/runner_test.go
ifireball: @yftacherzog please have a 2nd look, using testify now

yftacherzog: What's the rationale behind using a combination of assert and require?

ifireball: assert logs the error and keeps going while require exists the test immediately.

scripts/runner_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Run(tt.scriptFile, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+			got, err := LookPath(tt.scriptFile)
+			if tt.wantErr {
+				assert.Error(err)
+				return
+			}
+			require.Nil(err, "Must be able to find %v", tt.scriptFile)
```
yftacherzog, line 32: I think assert/require for `Error()` might be more straightforward and also follows the same path of the case that has error.
https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/assert#NoErrorf

yftacherzog (approved):
Looks great! Left one comment for you to consider.

yftacherzog:
lgtm

PR #25, opened 2023-07-27
test(RHTAPWATCH-383): segment-mass-uploader.sh

ifireball:
Tests for `segment-mass-uploader.sh`.

Along the way added a little statistics library that may be handy elsewhere.


stats/series.go
```diff
@@ -0,0 +1,77 @@
...
+	"golang.org/x/exp/constraints"
+)
+
+type Number interface {
+	constraints.Integer | constraints.Float
+}
+
+// Series calculates cumulative statistics on a series of numbers
+// For streamlined use, all operations are designed dot be error-free, results
+// my not always be useful though.
```
yftacherzog, line 16:
```suggestion
// may not always be useful though.
```

stats/series.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"golang.org/x/exp/constraints"
+)
+
+type Number interface {
+	constraints.Integer | constraints.Float
+}
+
+// Series calculates cumulative statistics on a series of numbers
+// For streamlined use, all operations are designed dot be error-free, results
```
yftacherzog, line 15: ?
```suggestion
// For streamlined use, all operations are designed to be error-free, results
```

stats/series.go
```diff
@@ -0,0 +1,77 @@
...
+func (s Series[T]) Total() T  { return s.total }
+
+func (s Series[T]) Avg() T {
+	if s.count <= 0 {
+		return 0
+	}
+	return s.total / T(s.count)
+}
+
+func (s Series[T]) Median() T {
```
yftacherzog, line 53: I think this should be midrange rather than median
```suggestion
func (s Series[T]) Midrange() T {
```

ifireball: will change to some other, more useful statistical property

ifireball: actually going to just drop it because I can't find any other useful statistic property that does not require me to store all the series data in order to calculate it.

stats/series_test.go
```diff
@@ -0,0 +1,57 @@
...
+			name: "3 samples",
+			samples: []int{2, 8, 8},
+			format: "Stats: %3d",
+			want: results{
+				len: 3,
+				min: 2,
+				max: 8,
+				total: 18,
+				avg: 6,
+				median: 5,
```
yftacherzog, line 38: can drop the median

stats/series_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIntSeries(t *testing.T) {
+	type results struct {
+		len uint
+		min, max, total, avg, median int
```
yftacherzog, line 13: can drop the median

yftacherzog (approved):

stats/series.go
```diff
@@ -0,0 +1,73 @@
...
+func (s Series[T]) Total() T  { return s.total }
+
+func (s Series[T]) Avg() T {
+	if s.count <= 0 {
+		return 0
+	}
+	return s.total / T(s.count)
+}
+
+func (s Series[T]) Format(f fmt.State, verb rune) {
```
Omeramsc, line 53: This function is pretty hard to understand. while the test makes it easier, I would add a doc comment.

stats/series_test.go
```diff
@@ -0,0 +1,56 @@
...
+			name: "3 samples",
+			samples: []int{2, 8, 8},
+			format: "Stats: %3d",
+			want: results{
+				len: 3,
+				min: 2,
+				max: 8,
+				total: 18,
+				avg: 6,
+				str: "Stats: min:   2 max:   8 avg:   6",
```
Omeramsc, line 38: Is there a reason we want 3 spaces here?

Omeramsc (commented):
isn't there a GO package for statistics that can fit what we need without reinventing the wheel? what are the benefits for us maintaining this code?

stats/series_test.go
ifireball: to test that the format string works.

ifireball:
> isn't there a GO package for statistics that can fit what we need without reinventing the wheel? what are the benefits for us maintaining this code?

I couldn't find one that was simple enough and supported using `int`s. Besides, its just min/max/total/avg calculation, most people would have not bothered to factor this out of the code that was using it.

stats/series.go
ifireball: check now

stats/series_test.go
Omeramsc: What I mean is, why do we want the format to have 3 spaces in the first place

ifireball: see how this helps to align output in the place where we use this as a part of the uploader test.

stats/series.go
Omeramsc: I guess? I worry more about the why than the what, meaning why do we want the formatting to be as it is in `rebuildFmtStr` or what can I expect the formatted string to look like

ifireball: @Omeramsc explaining what is the Formatter interface and how to implement it is IMO outside the scope of the code or the comments on it. 

`rebuildFmtStr` is basically rebuilding the format string that was passed to the `fmt` module function that is currently printing our object, since its only provided to us in its de-constructed form.

Whet this is doing is very simple: When you do `fmt.Printf("some '%' format expression", series_object")` it uses the format expression for formatting the numbers in the series object. This allows me to use everything the `fmt` package knows how to do for the numbers without inventing my own custom API for it.

Omeramsc (approved):
LGTM

PR #26, opened 2023-07-28
chore(RHTAPWATCH-173): Remove unecessary Application events

amisstea:
Only Application create events are desired at this time.

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -121,7 +121,7 @@ func GenApplicationQuery(index string) string {
 	return `search ` +
 		`index="` + index + `" ` +
 		`log_type=audit ` +
-		`NOT verb IN (get, watch, list, deletecollection) ` +
+		`verb=create ` +
 		`"responseStatus.code" IN (200, 201) ` +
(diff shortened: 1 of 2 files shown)


hmariset (approved):
LGTM

ifireball (commented):
Code-wise LGTM, but are we sure we only need Application creation events? 

In the UJ event lsts, application creation in the UI and picking a name are listed separately, are we sure both are a single event in the backend?

Omeramsc (approved):
LGTM, however I would add in the commit (or change in a separate commit) about all the lint changes. this could've been a much smaller PR

amisstea:
> LGTM, however I would add in the commit (or change in a separate commit) about all the lint changes. this could've been a much smaller PR

I broke this up into two commits rather than trying to alter the default `gofmt` execution in my IDE (vscode).

amisstea:
> Code-wise LGTM, but are we sure we only need Application creation events?
> 
> In the UJ event lists, application creation in the UI and picking a name are listed separately, are we sure both are a single event in the backend?

@ifireball yes, I've confirmed this. It's all part of the same backend event because it's the same as the resource name.

PR #27, opened 2023-07-31
test(RHTAPWATCH-382): test get-uid-map.sh

srivickynesh:
1. Reworked scripts from #20.
2. created sample test to test get-uid-map.sh returned output.
3. Modified kwok/user-signups.sh to patch complaintUsername to usersignup yaml.

How to test:
1. First build the image using `podman build --build-arg CLUSTER_NAME=tester -t kwok -f kwok/Dockerfile .`
2. Now run the go test

### DIFF
--- a/kwok/kwok_container_template.yml
+++ b/kwok/kwok_container_template.yml
@@ -1,0 +1,15 @@
+# Save the output of this file and use kubectl create -f to import
+# it into Kubernetes.
+#
+# Created with podman-4.5.1
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ .PodName}}
+spec:
+  containers:
+  - name: kwok
+    image: quay.io/srivickynesh/kwok:kwok-bundle
+    ports:
+    - containerPort: 8080
+      hostPort: 8080
--- a/kwok/user-signups.sh
+++ b/kwok/user-signups.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 
+total_users=9
 namespace="toolchain-host-operator"
+
 kwokctl --name "$CLUSTER_NAME" kubectl create namespace "$namespace"
 
-for i in $(seq 1 10)
+for i in $(seq 1 $total_users)
 do
 email="user$i@gmail.com"
 email_hash=$(printf "%s" "$email" | md5sum | awk '{ print $1 }')
@@ -26,4 +28,7 @@ spec:
   userid: "f:528d74ff-f703-47ed-9cd5-f0ct$i:user$i"
   username: "user$i"
 EOF
+
+kwokctl --name "$CLUSTER_NAME" kubectl patch usersignup "user$i" -n toolchain-host-operator --type=merge --patch "{\"status\":{\"compliantUsername\":\"user$i\"}}" --subresource=status
+
 done
--- a/kwok/Dockerfile
+++ b/kwok/Dockerfile
@@ -3,7 +3,8 @@ FROM registry.k8s.io/kwok/cluster:v0.3.0-k8s.v1.27.3
 ARG CLUSTER_NAME=kwok
 ENV CLUSTER_NAME=${CLUSTER_NAME}
 
-COPY --chmod=755 user-signups.sh entrypoint.sh toolchain.dev.openshift.com_usersignups.yaml ./
+COPY --chmod=755 kwok/user-signups.sh kwok/entrypoint.sh kwok/toolchain.dev.openshift.com_usersignups.yaml ./
+COPY --chown=root:root --chmod=755 scripts/* /usr/local/bin
 
 RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
     kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
--- a/kwok/getuidmap.go
+++ b/kwok/getuidmap.go
@@ -1,0 +1,97 @@
+package kwok
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+)
+
+const (
+	contextName      = "kwok"
+	clusterServerURL = "http://127.0.0.1:8080"
+	scriptPath       = "../scripts/get-uid-map.sh"
+
+	FailedToSetClusterMsg = "failed to set cluster: %w"
+	FailedToSetContextMsg = "failed to set context: %w"
+	FailedToUseContextMsg = "failed to use context: %w"
+)
+
+type ScriptExecutor interface {
+	Execute(scriptPath string) ([]byte, error)
+}
+
+type ShellScriptExecutor struct{}
+
+func (s *ShellScriptExecutor) Execute(scriptPath string) ([]byte, error) {
+	return exec.Command("/bin/sh", scriptPath).CombinedOutput()
+}
+
+func runOcCommand(args ...string) error {
+	cmd := exec.Command("oc", args...)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to run oc command: %w", err)
+	}
+	return nil
+}
+
+func setCluster(name, server string) error {
+	return runOcCommand("config", "set-cluster", name, "--server="+server)
+}
+
+func setContext(name, cluster string) error {
+	return runOcCommand("config", "set-context", name, "--cluster="+cluster)
+}
+
+func useContext(name string) error {
+	return runOcCommand("config", "use-context", name)
+}
+
+func ExecuteAndParseScript(executor ScriptExecutor, scriptPath string) (map[string]int64, error) {
+	m := make(map[string]int64)
+
+	output, err := executor.Execute(scriptPath)
+	if err != nil {
+		return m, fmt.Errorf("failed to execute command\nOutput: %s\nError: %w", string(output), err)
+	}
+
+	if err := json.Unmarshal(output, &m); err != nil {
+		return m, fmt.Errorf("failed to parse result: %w", err)
+	}
+
+	return m, nil
+}
+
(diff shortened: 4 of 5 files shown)


scripts/get-uid-map.sh
```diff
@@ -2,12 +2,22 @@
 # get-uid-map.sh
 #   Read UserSignup resources and generate a JSON object mapping from cluster
 #   usernames to SSO UDIs.
-#   This script assumes `oc` is preconfigured to connect to the right cluster
+#   This script assumes `oc` or 'kwokctl --name "$CLUSTER_NAME" kubectl' is preconfigured to connect to the right cluster
```
yftacherzog, line 5:
```suggestion
#   This script assumes `oc` or 'kwokctl --name "$CLUSTER_NAME" kubectl' are preconfigured to connect to the right cluster
```

kwok/tests/usersignup.go
yftacherzog, line 1: The name and location for this file are not obvious to me. What's the reason for it to have the same name as the shell script outside the `tests` directory? What's the purpose of this file? Is that a test file? Is it only used by `usersignup_test.go`?  If those are just helper functions for those tests, maybe they should be defined there?

kwok/tests/usersignup.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func PodmanBuildImage() (string, error) {
+	buildCmd := exec.Command("podman", "build", "--build-arg", "CLUSTER_NAME=test", "-t", "my-kwokctl-oc-image", "-f", "../Dockerfile", "../../.")
+	output, err := buildCmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("failed to build image: %w", err)
+	}
+
+	fmt.Printf("BuildImage output: %s\n", output)
```
yftacherzog, line 17: Do we want to keep this print?

kwok/tests/usersignup.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to start command: %w", err)
+	}
+
+	containerID := strings.TrimSpace(string(output))
+
+	return containerID, nil
+}
+
+func PodmanExecCommand(containerID string, command string) (map[string]int64, error) {
```
yftacherzog, line 36: I think it might be worth documenting that the provided command must be one that is expected to return json output.

kwok/tests/usersignup_test.go
```diff
@@ -0,0 +1,47 @@
...
+	if !validateMap(result) {
+		t.Errorf("Returned map was incorrect, got: %v", result)
+	}
+	log.Println("All tests Successful")
+}
+
+func validateMap(m map[string]int64) bool {
+	for i := 1; i <= 9; i++ {
+		user := fmt.Sprintf("user%d", i)
+		if val, ok := m[user]; !ok || int(val) < 52542471 || int(val) > 52542479 {
```
yftacherzog, line 42: What are those numbers? Can we make them constants of some sort to make the test more obvious?

kwok/tests/usersignup_test.go
```diff
@@ -0,0 +1,47 @@
...
+	log.Println("Result of get-uid-map.sh: ", result)
+
+	if !validateMap(result) {
+		t.Errorf("Returned map was incorrect, got: %v", result)
+	}
+	log.Println("All tests Successful")
+}
+
+func validateMap(m map[string]int64) bool {
+	for i := 1; i <= 9; i++ {
```
yftacherzog, line 40: I have some reservations regarding the fact that we have those 1-9 users hard coded here, as it couples it to some hard coded stuff done elsewhere.
Does `get-uid-map.sh` returns any other users (with values outside the range we test for)? Can we run the validation on all key-values in its result instead of having to iterate those specific users?

kwok/tests/usersignup.go
srivickynesh: thats the sh script that creates usersignups and the tests have tests to validate them

kwok/tests/usersignup_test.go
srivickynesh: Noted

kwok/tests/usersignup.go
```diff
@@ -0,0 +1,65 @@
...
+		return fmt.Errorf("error stopping the container: %w", err)
+	}
+
+	cmd = exec.Command("podman", "rm", "-f", containerID)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error removing the container: %w", err)
+	}
+
+	return nil
+}
```
Omeramsc, line 65: Are we gonna use this file or is it temporary? it mostly iterates over what's done in my patch

srivickynesh: I'm planning to use it once it is merged. had that here as a temp way to test things.

yftacherzog (approved):

kwok/tests/usersignup.go
Omeramsc: In the meanwhile you can rebase on top of my change. 
You can do so after this change is merged, saving some time, but it just adds a lot of code to review here which soon won't be needed

srivickynesh: I will import podman kube play lib from container fixtures but that needs some modification in existing. But that's not a lot.

kwok/tests/kwok-playbook.yml
```diff
@@ -0,0 +1,17 @@
+# Save the output of this file and use kubectl create -f to import
```
Omeramsc, line 1: Barak suggested that `playbook` is not a good name, I tend to agree (the existence of Ansible playbooks is confusing). I've gone with manifest instead, but do tell if you have better suggestions.

kwok/tests/usersignup_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}), "Map with an empty key should be rejected")
+	})
+
+	t.Run("no value key", func(t *testing.T) {
+		assert.False(validateMap(map[string]int64{
+			"<no value>": 87654321,
+		}), "Map with '<no value>' key should be rejected")
+	})
+
+	t.Run("empty map", func(t *testing.T) {
```
Omeramsc, line 60: I see multiple test-cases for validateMap, why not use the parametrize tests (want/got with a struct of test cases)?

Omeramsc (changes requested):
@srivickynesh I rather you rebase your patch on top of my commit, will be easier for you to pull any changes I do on my PR.

ifireball:
Since #20 is not merged yet, please help @Omeramsc get things right there instead of modifying it again here.

kwok/usersignup_test.go
```diff
@@ -0,0 +1,88 @@
+package kwok
```
Omeramsc, line 1: The name of the file suggest it is testing `useContext`, `setContext`, and `setCluster`; but as the title of the PR suggest, it is testing `get-uid-map` using the functions from `usersignup` as pre-configurations. Consider changing it, as it creates confusion.

kwok/usersignup.go
```diff
@@ -0,0 +1,30 @@
+package kwok
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func setCluster(name, server string) error {
```
Omeramsc, line 8: are those used only for the tests, or with out code? if the former, might consider include them as fixtures in the test file.

kwok/user-signups.sh
```diff
@@ -3,7 +3,7 @@
 namespace="toolchain-host-operator"
 kwokctl --name "$CLUSTER_NAME" kubectl create namespace "$namespace"
 
-for i in $(seq 1 10)
+for i in $(seq 1 9)
```
Omeramsc, line 6: What's the reason for using `9` specifically? might be better to include it as a named variable.

Omeramsc (changes requested):

kwok/user-signups.sh
srivickynesh: I am also setting up sso-id with this.

kwok/kwok_container_template.yml
```diff
@@ -0,0 +1,15 @@
...
+kind: Pod
+metadata:
+  name: {{ .PodName}}
+spec:
+  containers:
+  - name: kwok
+    image: localhost/kwok:latest
+    ports:
+    - containerPort: 8080
+      hostPort: 8080
```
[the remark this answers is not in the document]
srivickynesh: yes, Else I would need to make changes to buildAndRunPod()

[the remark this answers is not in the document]
Omeramsc: We can set a specific pod-name and ignore the templating operation (as we do with the ports)
but besides making the it clear that we don't need to run it concurrently, there are no other benefits, so it can stay as-is...

kwok/user-signups.sh
Omeramsc: all the more reason to include it as a named variable.

kwok/usersignup_test.go
Omeramsc: better. is there a reason not to name it `getuidmap_test` or something similar? is it expected to test more things in the future?

srivickynesh: It runs the get-uid-map script from the script to gather the user signups and passes them as input for user signup tests. Ideally, both are correct.

kwok/kwok.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+)
+
+const (
+	contextName      = "kwok"
+	clusterServerURL = "http://127.0.0.1:8080"
+	scriptPath       = "../scripts/get-uid-map.sh"
```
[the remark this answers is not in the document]
srivickynesh: Yep on 8080.

Omeramsc (approved):
LGTM

hmariset (approved):
LGTM

amisstea (approved):

PR #28, opened 2023-08-01
Querygen refactor

ifireball:
Re-factor all the query generation code so that its not almost impossible to generate invalid queries.


querygen/fieldset.go
```diff
@@ -0,0 +1,181 @@
...
+// then in the output
+//
+// The keys in the FieldSet are the output field names while the values include
+// details about how to obtain the output values and how to present them.
+//
+// The zero value for the value struct defines that the output field is copied
+// as-is from the input.
+type FieldSet map[string]struct {
+	// srcFields defines which input fields to use. Each field in the list is
+	// used as a fallback for when all the fields the precede it are NULL.
```
yftacherzog, line 26:
```suggestion
	// used as a fallback for when all the fields that precede it are NULL.
```

amisstea (approved):
LGTM. Nice job!

PR #29, opened 2023-08-03
fix: image build failing in OpenShift & Podman

ifireball:
The main image build was failing when running in OpenShift and Podman because it was being done in a root-owned directory.


yftacherzog (approved):

srivickynesh (approved):

Omeramsc (approved):
LGTM

PR #30, opened 2023-08-04
RHTAPWATCH-154: Produce a segment event when a build PipelineRun starts

amisstea:
A `PipelineRun` started event is detected by searching for update events on the `status` subresource where the `Succeeded` condition has a reason of `Running` with 0 tasks complete.

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -7,12 +7,12 @@ package querygen
 func GenApplicationQuery(index string) string {
 	q, _ := UserJourneyQueryGen(
 		index,
-		`verb=create ` +
-		`"responseStatus.code" IN (200, 201) ` +
-		`"objectRef.apiGroup"="appstudio.redhat.com" ` +
-		`"objectRef.resource"="applications" ` +
-		`("impersonatedUser.username"="*" OR (user.username="*" AND NOT user.username="system:*")) ` +
-		`(verb!=create OR "responseObject.metadata.resourceVersion"="*")`,
+		`verb=create `+
+			`"responseStatus.code" IN (200, 201) `+
+			`"objectRef.apiGroup"="appstudio.redhat.com" `+
+			`"objectRef.resource"="applications" `+
+			`("impersonatedUser.username"="*" OR (user.username="*" AND NOT user.username="system:*")) `+
+			`(verb!=create OR "responseObject.metadata.resourceVersion"="*")`,
 		[]string{"name", "userId"},
 	)
 	return q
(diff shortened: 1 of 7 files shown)


querygen/querygen.go
```diff
@@ -7,29 +7,53 @@ package querygen
...
+		index,
+		`verb=update `+
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="pipelineruns" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.startTime"="*" `+
+			`| eval event_subject="Build PipelineRun", event_verb="started"`+
```
ifireball, line 51: might as well pass-through the complete `event` string instead of hacking subject and verb like this.

this will remove the need to modify the field definitions for them.

scripts/fetch-uj-records.sh
```diff
@@ -56,4 +56,4 @@ $QUERYGEN -0 --index="$SPLUNK_INDEX" \
     --data output_mode=json \
     --data earliest_time="$QUERY_EARLIEST_TIME" \
     --data latest_time="$QUERY_LATEST_TIME" \
-    --data search=Q
+    --data-urlencode search=Q
```
ifireball, line 59: I wonder why we didn't need this before. Also I think this is sending POST requests, so IMO shouldn't need to URL-encode.

querygen/querygen.go
amisstea: Changed. I did have to create a field definition for `event` and some simple fallback logic in `splunk-to-segment.sh` to prioritize using this field over the combination of `event_subject` and `event_verb`.

scripts/fetch-uj-records.sh
amisstea: It's specifically needed to send a `%` char in the splunk query (see the commit description). This use case also happens to be well documented by Splunk: https://docs.splunk.com/Documentation/Splunk/9.1.0/RESTUM/RESTusing

> In most instances, the -d parameter will sufficiently URI-encode a search string. However, if you are having problems logging in to an endpoint or running a search, use the --data-urlencode parameter instead of -d to ensure all characters in the string are URI-encoded.

querygen/querygen.go
ifireball: Didn't we have this already? I remember making an issue for it...

ifireball (approved):
/lgtm

querygen/querygen.go
amisstea: Perhaps you're thinking of [RHTAPWATCH-373](https://issues.redhat.com/browse/RHTAPWATCH-373)? Some of the AC in [RHTAPWATCH-384](https://issues.redhat.com/browse/RHTAPWATCH-384) touches upon this behavior too.

PR #31, opened 2023-08-06
chore(RHTAPWATCH-361): simple Splunk test

Omeramsc:
A test on a Splunk query to demonstrate GO test-framework. It's running the fake cluster and checks that the count of logs indexed for the test_index is as we expect.

depends on #20

### DIFF
--- a/splunk/splunk.go
+++ b/splunk/splunk.go
@@ -1,0 +1,27 @@
+package splunk
+
+import (
+	"fmt"
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/redhat-appstudio/segment-bridge.git/containerfixture"
+)
+
+func getRecordsCount(host, port, owner, app, index string) (string, error) {
+	login := fmt.Sprintf("%s:%s", containerfixture.Username, containerfixture.Password)
+	endpoint := fmt.Sprintf(
+		"https://%s:%s/servicesNS/%s/%s/search/v2/jobs/export?output_mode=csv",
+		host, port, owner, app,
+	)
+	searchQuery := "search=" + fmt.Sprintf("search index=%s | stats count", index)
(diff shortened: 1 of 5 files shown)


splunk/splunk.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/redhat-appstudio/segment-bridge.git/containerfixture"
+)
+
+func getRecordsCount(host, port, owner, app, index string) (string, error) {
+	login := fmt.Sprintf("%s:%s", containerfixture.Username, containerfixture.Password)
+	endpoint := fmt.Sprintf(
+		"https://%s:%s/servicesNS/%s/%s/search/v2/jobs/export?output_mode=csv",
+		host, port, owner, app,
+	)
+	searchQuery := "search=" + (fmt.Sprintf("search index=%s | stats count", index))
```
yftacherzog, line 18: Do we need the outer parentheses here?

Omeramsc: Nope, good catch.

yftacherzog (approved):

ifireball (changes requested):
Not looking at the code here yet, I think this can be a part of #20, so that well have some tests included for the code there.

Omeramsc:
@b

> Not looking at the code here yet, I think this can be a part of #20, so that well have some tests included for the code there.

I get the value of having the tests with the framework, but I separated them on purpose. #20 is big enough and other changes depends on it, adding GO test code (which may need multiple changes) will prolong it even further.

ifireball:
> > Not looking at the code here yet, I think this can be a part of #20, so that well have some tests included for the code there.
> 
> I get the value of having the tests with the framework, but I separated them on purpose. #20 is big enough and other changes depends on it, adding GO test code (which may need multiple changes) will prolong it even further.

Are we cheating ourselves? You can't reliably merge that code without having this code - there are both parts of the same work - if you really need so separation use different commits within the same PR.

PR #34, opened 2023-08-15
chore(RHTAPWATCH-151): build PipelineRun success and failure events

ShakedAviv50313:
Produce a segment event when a build PipelineRun succeeds or fails.

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -87,3 +87,27 @@ func GenClairScanCompletedQuery(index string) string {
 	)
 	return q
 }
+
+// GenBuildPipelineRunCompletedQuery returns a Splunk query for generating Segment events
+// representing success or failure of AppStudio build PipelineRuns.
+func GenBuildPipelineRunCompletedQuery(index string) string {
+	q, _ := UserJourneyQueryGen(
+		index,
+		`verb=update `+
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="pipelineruns" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath responseObject.status.conditions{}`+
+			`| mvexpand responseObject.status.conditions{}`+
+			`| search responseObject.status.conditions{}.type="Succeeded" AND ` +
+			`(responseObject.status.conditions{}.reason="Completed" OR `+
+			`responseObject.status.conditions{}.reason="Failed")`+
+			`| eval event="Build PipelineRun ".mvindex('responseObject.status.conditions{}.reason', 0)`,
+		[]string{"namespace", "application", "component", "status_message", "status_reason"},
+	)
+	return q
+}
--- a/querygen/querygen_test.go
+++ b/querygen/querygen_test.go
@@ -29,3 +29,8 @@ func TestGenClairScanCompletedQuery(t *testing.T) {
 	out := GenClairScanCompletedQuery("some_index")
 	assert.NotEqual(t, "", out)
 }
+
+func TestGenBuildPipelineRunCompletedQuery(t *testing.T) {
+	out := GenBuildPipelineRunCompletedQuery("some_index")
+	assert.NotEqual(t, "", out)
+}
--- a/cmd/querygen/main.go
+++ b/cmd/querygen/main.go
@@ -58,5 +58,9 @@ func main() {
 			Title: "Clair scan TaskRun completion events",
 			Query: querygen.GenClairScanCompletedQuery(*index),
 		},
+		{
+			Title: "Build PipelineRun Completed or Failed event",
+			Query: querygen.GenBuildPipelineRunCompletedQuery(*index),
+		},
 	}))
 }
--- a/querygen/userjourney_fs.go
+++ b/querygen/userjourney_fs.go
@@ -43,6 +43,14 @@ var UserJourneyFieldSet = FieldSet{
 		subObj:    "properties",
 		srcFields: []string{"clair_scan_result.vulnerabilities.low"},
 	},
+	"status_message": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.message"},
+	},
+	"status_reason": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.reason"},
+	},
 }
 
 var UserJourneyCommonFields = [...]string{


ShakedAviv50313:
This is just a draft, I will modify it tomorrow, no need to review this yet

querygen/querygen.go
```diff
@@ -57,3 +57,50 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+		index,
+		`verb=update `+
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="pipelineruns" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.startTime"="*" `+
+			`| eval event="Build PipelineRun Completed" `+
```
ifireball, line 74: Looks like you could build the "eval" string using the "reason" field you use in the `where` clause below, and therefore unit these two functions into one.

ifireball (changes requested):

querygen/querygen.go
```diff
@@ -57,3 +57,26 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="pipelineruns" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.startTime"="*" `+
+			`| spath path=responseObject.status.conditions{} output=conditions `+
+			`| mvexpand conditions `+
+			`| spath input=conditions `+
```
ifireball, line 76: Since this is very similar to the existing query in GenBuildPipelineRunStartedQuery, could both be combined?

The trouble with `spath` is that you can't see the fields that are being generated in the query string - so things are less readable. Perhaps we could give up on the first `spath` and run `mvexpand` directly on the nested field?

As for the 2nd `spath` could we do without it? As far as I can tell, it would make the "where" expression a bit longer and we would need an eval to generate "reason" and "message" which would be clearer IMO.

Also consider using "search" rather then where to keep things consistent - as long as you don't need any eval functions, search should be sufficient.

querygen/querygen_test.go
```diff
@@ -20,7 +20,8 @@ func TestGenBuildPipelineRunCreatedQuery(t *testing.T) {
 	assert.NotEqual(t, "", out)
 }
 
-func TestGenBuildPipelineRunStartedQuery(t *testing.T) {
-	out := GenBuildPipelineRunStartedQuery("some_index")
+func TestGenBuildPipelineRunCompletedQuery(t *testing.T) {
```
ifireball, line 23: As far as I can tell we still need a test for `GenBuildPipelineRunStartedQuery` because we still use that function

ifireball (changes requested):

querygen/querygen.go
amisstea: I made some suggestions on https://github.com/redhat-appstudio/segment-bridge/pull/36#discussion_r1304940201 related to these comments. PTAL and maybe we can incorporate them here.

ShakedAviv50313:
i omitted the second `spath` and replaced `where` with `search`, the first spath is crucial as using mvexpand directly at responseObject.status.conditions{} without the spath returns:
``` 27 of 48 peers omitted, Field 'responseObject.status.conditions{}' does not exist in the data. ```

File: querygen/querygen_test.go (hunk 3 above)
Comment (ShakedAviv50313) [reply querygen/querygen_test.go:23]: Done

**Comment by ShakedAviv50313** [comment]
@ifireball @amisstea WDYT about the changes?

File: querygen/querygen_test.go (hunk 3 above)
Comment (amisstea) [reply querygen/querygen_test.go:23]: @ShakedAviv50313 we're now missing `TestGenBuildPipelineRunCreatedQuery`. Please ensure we have 3 tests for build PipelineRuns and that they call their respective query generation funcs.

File: querygen/querygen.go (hunk 1 above)
Comment (amisstea) [reply querygen/querygen.go]: @ShakedAviv50313 have you attempted this?

cc @Kousalya1998 as it relates to #36

**Review by amisstea — CHANGES_REQUESTED** [review]

File: querygen/querygen.go [hunk 4]
```diff
@@ -57,3 +57,26 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+		index,
+		`verb=update `+
+		`"responseStatus.code"=200 `+
+		`"objectRef.apiGroup"="tekton.dev" `+
+		`"objectRef.resource"="pipelineruns" `+
+		`"objectRef.subresource"="status" `+
+		`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+		`"responseObject.metadata.resourceVersion"="*" `+
+		`"responseObject.status.startTime"="*" `+
+		`| spath responseObject.status.conditions{}.reason output=reason` +
```
amisstea, line 74: We should be able to remove this `spath` command by using `responseObject.status.conditions{}.reason` instead of `reason`.

querygen/querygen.go
```diff
@@ -57,3 +57,26 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+	q, _ := UserJourneyQueryGen(
+		index,
+		`verb=update `+
+		`"responseStatus.code"=200 `+
+		`"objectRef.apiGroup"="tekton.dev" `+
+		`"objectRef.resource"="pipelineruns" `+
+		`"objectRef.subresource"="status" `+
+		`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+		`"responseObject.metadata.resourceVersion"="*" `+
+		`"responseObject.status.startTime"="*" `+
```
amisstea, line 73: This should filter by any defined completion time, not start time.

amisstea (changes requested):
@ShakedAviv50313 please rebase your branch and resolve merge conflicts.

querygen/querygen.go
amisstea: To do this I think you'll need to "dedupe" the value like the FieldSet code does. E.g.
```
eval event="Build PipelineRun ".mvindex('responseObject.status.conditions{}.reason', 0)
```

querygen/querygen_test.go
ShakedAviv50313: Done, I rebased this branch with upstream/main

amisstea (approved):

ShakedAviv50313:
@ifireball

querygen/querygen.go
```diff
@@ -87,3 +87,25 @@ func GenClairScanCompletedQuery(index string) string {
...
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="pipelineruns" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.labels.pipelines.appstudio.openshift.io/type"=build `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath responseObject.status.conditions{}`+
+			`| mvexpand responseObject.status.conditions{}`+
+			`| search responseObject.status.conditions{}.type="Succeeded" AND (responseObject.status.conditions{}.reason="Completed" OR responseObject.status.conditions{}.reason="Failed")`+
```
ifireball, line 106: please break this long string

ifireball (approved):
Approving with the understanding that the code is going to be improved with work by @Kousalya1998

ShakedAviv50313:
Last changes have been made, (breaking the long string).
I cannot merge this PR because I do not have write access to the repo

PR #36, opened 2023-08-17
chore(RHTAPWATCH-95): Produce a segment event when the clair-scan task completes

Kousalya1998:
Produce a segment event when the clair-scan task completes

Task:
[RHTAPWATCH-95](https://issues.redhat.com/browse/RHTAPWATCH-95)

### DIFF
--- a/cmd/querygen/main.go
+++ b/cmd/querygen/main.go
@@ -54,5 +54,9 @@ func main() {
 			Title: "Build PipelineRun started events",
 			Query: querygen.GenBuildPipelineRunStartedQuery(*index),
 		},
+		{
+			Title: "Clair scan TaskRun completion events",
+			Query: querygen.GenClairScanCompletedQuery(*index),
+		},
 	}))
 }
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -57,3 +57,33 @@ func GenBuildPipelineRunStartedQuery(index string) string {
 	)
 	return q
 }
+
+// GenClairScanCompletedQuery returns a Splunk query for generating Segment events
+// when the clair-scan task completes.
+func GenClairScanCompletedQuery(index string) string {
+	q, _ := UserJourneyQueryGen(
+		index,
+		`verb=update `+
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="tekton.dev" `+
+			`"objectRef.resource"="taskruns" `+
+			`"objectRef.subresource"="status" `+
+			`"requestObject.metadata.labels.tekton.dev/pipelineTask"="clair-scan" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath responseObject.status.conditions{} `+
+			`| mvexpand responseObject.status.conditions{} `+
+			`| search responseObject.status.conditions{}.type="Succeeded" `+
+			`responseObject.status.conditions{}.reason="Succeeded" `+
+			`responseObject.status.conditions{}.status="True" `+
+			`| spath responseObject.status.taskResults{} `+
+			`| mvexpand responseObject.status.taskResults{} `+
+			`| search responseObject.status.taskResults{}.name="CLAIR_SCAN_RESULT" `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.critical output=clair_scan_result.vulnerabilities.critical `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.high output=clair_scan_result.vulnerabilities.high `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.medium output=clair_scan_result.vulnerabilities.medium `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.low output=clair_scan_result.vulnerabilities.low `+
+			`| eval event="Clair scan TaskRun completed"`,
+		[]string{"namespace", "application", "component", "vulnerabilities_critical", "vulnerabilities_high", "vulnerabilities_medium", "vulnerabilities_low"},
+	)
+	return q
+}
--- a/querygen/userjourney_fs.go
+++ b/querygen/userjourney_fs.go
@@ -27,6 +27,22 @@ var UserJourneyFieldSet = FieldSet{
 		subObj:    "properties",
 		srcFields: []string{"responseObject.metadata.labels.appstudio.openshift.io/component"},
 	},
+	"vulnerabilities_critical": {
+		subObj:    "properties",
+		srcFields: []string{"clair_scan_result.vulnerabilities.critical"},
+	},
+	"vulnerabilities_high": {
+		subObj:    "properties",
+		srcFields: []string{"clair_scan_result.vulnerabilities.high"},
+	},
+	"vulnerabilities_medium": {
+		subObj:    "properties",
+		srcFields: []string{"clair_scan_result.vulnerabilities.medium"},
+	},
+	"vulnerabilities_low": {
+		subObj:    "properties",
+		srcFields: []string{"clair_scan_result.vulnerabilities.low"},
+	},
 }
 
 var UserJourneyCommonFields = [...]string{
--- a/querygen/querygen_test.go
+++ b/querygen/querygen_test.go
@@ -24,3 +24,8 @@ func TestGenBuildPipelineRunStartedQuery(t *testing.T) {
 	out := GenBuildPipelineRunStartedQuery("some_index")
 	assert.NotEqual(t, "", out)
 }
+
+func TestGenClairScanCompletedQuery(t *testing.T) {
+	out := GenClairScanCompletedQuery("some_index")
+	assert.NotEqual(t, "", out)
+}


svatares (approved):
LGTM

cmd/querygen/main.go
```diff
@@ -54,5 +54,9 @@ func main() {
 			Title: "Build PipelineRun started events",
 			Query: querygen.GenBuildPipelineRunStartedQuery(*index),
 		},
+		{
+			Title: "Build TaskRun Completed event",
```
Omeramsc, line 58: Isn't the event about a `clair-scan`?  is that the only type of TaskRun?

cmd/querygen/main.go
```diff
@@ -54,5 +54,9 @@ func main() {
...
+			Query: querygen.GenBuildTaskRunCompletedQuery(*index),
```
Omeramsc, line 59: same

querygen/querygen.go
```diff
@@ -57,3 +57,32 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+			`| spath input=conditions `+
+			`| where type="Succeeded" AND reason="Succeeded" AND status="True" `+
+			`| spath path=responseObject.status.taskResults{} output=taskResults `+
+			`| mvexpand taskResults `+
+			`| where name="CLAIR_SCAN_RESULT" `+
+			`| eval vulnerabilities_critical=mvindex('responseObject.status.taskResults{}.value.vulnerabilities.critical', 0),
+					vulnerabilities_high=mvindex('responseObject.status.taskResults{}.value.vulnerabilities.high', 0),
+					vulnerabilities_medium=mvindex('responseObject.status.taskResults{}.value.vulnerabilities.medium', 0),
+					vulnerabilities_low=mvindex('responseObject.status.taskResults{}.value.vulnerabilities.low', 0) `+
+			`| fields vulnerabilities_critical, vulnerabilities_high, vulnerabilities_medium, vulnerabilities_low `,
```
amisstea, line 84: We shouldn't declare fields like this. Instead, take a look at the `UserJourneyFieldSet` structure in `querygen/userjourney_fs.go` for how to do this part.

cmd/querygen/main.go
amisstea:
```suggestion
			Query: querygen.GenClairScanCompletedQuery(*index),
```

cmd/querygen/main.go
amisstea:
```suggestion
			Title: "Clair scan TaskRun completion events",
```

amisstea (changes requested):

querygen/querygen.go
Kousalya1998: @amisstea I followed `UserJourneyFieldSet` structure and created the respected fields. PLAT and let me know your suggestions.

querygen/querygen.go
```diff
@@ -57,3 +57,32 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+			`"objectRef.subresource"="status" `+
+			`"requestObject.metadata.labels.tekton.dev/pipelineTask"="clair-scan" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath path=responseObject.status.conditions{} output=conditions `+
+			`| mvexpand conditions `+
+			`| spath input=conditions `+
+			`| where type="Succeeded" AND reason="Succeeded" AND status="True" `+
+			`| spath path=responseObject.status.taskResults{} output=taskResults `+
+			`| mvexpand taskResults `+
+			`| where name="CLAIR_SCAN_RESULT" `+
```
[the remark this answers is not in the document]
Kousalya1998: @amisstea Now it should show `45 events` for the last `24 hours` of data.

hmariset (approved):
LGTM

querygen/querygen.go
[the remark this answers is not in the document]
ifireball: do we really need 8 different splunk commands here? could we simplify things down somehow?

is there some other way to check for some value in some array within the event record other then blowing it up into multiple records via `mvexpend`?

also I commented elsewhere already that:

1.  I'd rather see `search` being used instead of `where` for consistency
2. It rather see `eval` being used instead of `spath` because it makes the result fields being defined explicit.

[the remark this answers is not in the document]
amisstea: FWIW, while working on the build pipelinerun started events query I didn't find an alternative to `mvexpand` for working with nested arrays but perhaps I didn't look closely enough. @Kousalya1998 please be on the lookout for a better alternative, if possible.

As for using `eval` over `spath` generally, good idea. Some of these nested field names are too generic (e.g. `name`, `value`, etc.). But I think we may need one `spath` command to parse the JSON string value from the `value` field of the task result. This is currently omitted and probably part of the reason this query doesn't return proper results yet.

querygen/querygen.go
```diff
@@ -57,3 +57,33 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+			`| where type="Succeeded" AND reason="Succeeded" AND status="True" `+
+			`| spath path=responseObject.status.taskResults{} output=taskResults `+
+			`| mvexpand taskResults `+
+			`| spath input=taskResults `+
+			`| search name="CLAIR_SCAN_RESULT" `+
+			`| spath input=value `+
+			`| eval vulnerabilities_critical=mvindex('vulnerabilities.critical', 0),
+					vulnerabilities_high=mvindex('vulnerabilities.high', 0),
+					vulnerabilities_medium=mvindex('vulnerabilities.medium', 0),
+					vulnerabilities_low=mvindex('vulnerabilities.low', 0) `,
```
amisstea, line 85: Here's a version that's:

1. Much more explicit (at the expense of some repetition). There are no more auto-extracted fields using `spath`.
2. Much safer and more consistent with field naming (e.g. `responseObject.status.conditions{}.type` vs `type`).
3. Uses `search` instead of `where`.
4. Specifies the `event` field to bypass the default event naming behavior in the `splunk-to-segment.sh` script.

```suggestion
			`| spath responseObject.status.conditions{} `+
			`| mvexpand responseObject.status.conditions{} `+
			`| search responseObject.status.conditions{}.type="Succeeded" `+
			`responseObject.status.conditions{}.reason="Succeeded" `+
			`responseObject.status.conditions{}.status="True" `+
			`| spath responseObject.status.taskResults{} `+
			`| mvexpand responseObject.status.taskResults{} `+
			`| search responseObject.status.taskResults{}.name="CLAIR_SCAN_RESULT" `+
			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.critical output=clair_scan_result.vulnerabilities.critical `+
			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.high output=clair_scan_result.vulnerabilities.high `+
			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.medium output=clair_scan_result.vulnerabilities.medium `+
			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.low output=clair_scan_result.vulnerabilities.low `+
			`| eval event="Clair scan TaskRun completed"`,
```

This necessitates a change to the `srcFields` in the fieldset as well.

amisstea (changes requested):

querygen/querygen.go
Kousalya1998: @amisstea I have now updated the query with your suggestions and compared the result. It works.

querygen/userjourney_fs.go
```diff
@@ -27,6 +27,10 @@ var UserJourneyFieldSet = FieldSet{
 		subObj:    "properties",
 		srcFields: []string{"responseObject.metadata.labels.appstudio.openshift.io/component"},
 	},
+	"vulnerabilities_critical": {subObj: "properties", srcFields: []string{"vulnerabilities.critical"}},
+	"vulnerabilities_high":     {subObj: "properties", srcFields: []string{"vulnerabilities.high"}},
+	"vulnerabilities_medium":   {subObj: "properties", srcFields: []string{"vulnerabilities.medium"}},
+	"vulnerabilities_low":      {subObj: "properties", srcFields: []string{"vulnerabilities.low"}},
```
amisstea, line 33: These `srcFields` also need updating to match the new field names. E.g. `clair_scan_result.vulnerabilities.critical`.

Be sure to run the `fetch-uj-records.sh` script to verify that all the expected fields are populated for this query.

querygen/querygen.go
```diff
@@ -57,3 +57,33 @@ func GenBuildPipelineRunStartedQuery(index string) string {
...
+			`| spath responseObject.status.taskResults{} `+
+			`| mvexpand responseObject.status.taskResults{} `+
+			`| search responseObject.status.taskResults{}.name="CLAIR_SCAN_RESULT" `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.critical output=clair_scan_result.vulnerabilities.critical `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.high output=clair_scan_result.vulnerabilities.high `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.medium output=clair_scan_result.vulnerabilities.medium `+
+			`| spath input=responseObject.status.taskResults{}.value path=vulnerabilities.low output=clair_scan_result.vulnerabilities.low `+
+			`| eval event="Clair scan TaskRun completed"`,
+		[]string{"namespace", "application", "component", "vulnerabilities_critical", "vulnerabilities_high", "vulnerabilities_medium", "vulnerabilities_low"},
+	)
```
ifireball, line 87: This result is not good - this is a hugs messy and unreadable string.

could we generate some of those fields via our existing FiedlSet mechanism?

I understand we may have a need to run search expressions after having generated some field values, we may need to expand FieldSet a bit to support this.

Mixing field value generation in what is supposed to only be a predicate string, while I thought to be tolerable at first, is not longer looking to me like it would be sustainable approach.

Specifically WRT the `spath` statements, it seems `eval` has an [spath function][1] that could be used in conjunction with our existing `FieldSet` code (If this become commonly used, I would, however, add a field to the FieldSet struct to simplify this case.

[1]: https://docs.splunk.com/Documentation/Splunk/9.1.0/SearchReference/TextFunctions#spath.28.26lt.3Bvalue.26gt.3B.2C.26lt.3Bpath.26gt.3B.29

querygen/userjourney_fs.go
```diff
@@ -27,6 +27,10 @@ var UserJourneyFieldSet = FieldSet{
 		subObj:    "properties",
 		srcFields: []string{"responseObject.metadata.labels.appstudio.openshift.io/component"},
 	},
+	"vulnerabilities_critical": {subObj: "properties", srcFields: []string{"clair_scan_result.vulnerabilities.critical"}},
```
ifireball, line 30: Try keeping the line length in the code below 100 characters or so so GitHub does not end up word-wrapping the code.

Kousalya1998: @ifireball I have updated the code. Ty.

PR #39, opened 2023-08-31
chore(RHTAPWATCH-183): Produce a segment event when a Release succeeds or fails.

Kousalya1998:
# Description
Produce a segment event when a Release succeeds or fails.
- Ensure that Release resource events are Sent to Segment
- Ensure that the resource success/failure state changes are reflected to Segment.

[RHTAPWATCH-183](https://issues.redhat.com/browse/RHTAPWATCH-183)

## How Has This Been Tested?
Tested the query locally in Splunk.

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -106,7 +106,7 @@ func GenBuildPipelineRunCompletedQuery(index string) string {
 			`"responseObject.status.completionTime"="*" `+
 			`| spath responseObject.status.conditions{}`+
 			`| mvexpand responseObject.status.conditions{}`+
-			`| search responseObject.status.conditions{}.type="Succeeded" AND ` +
+			`| search responseObject.status.conditions{}.type="Succeeded" AND `+
 			`(responseObject.status.conditions{}.reason="Completed" OR `+
 			`responseObject.status.conditions{}.reason="Failed")`+
 			`| eval event="Build PipelineRun ".mvindex('responseObject.status.conditions{}.reason', 0)`,
@@ -115,6 +115,28 @@ func GenBuildPipelineRunCompletedQuery(index string) string {
 			"status_message", "status_reason",
 			"repo", "commit_sha", "target_branch",
 		},
+	)
+	return q
+}
+
+// GenReleaseCompletedQuery returns a Splunk query for generating Segment events
+// representing the Release resource success/failure state changes.
+func GenReleaseCompletedQuery(index string) string {
+	q, _ := UserJourneyQueryGen(
+		index,
+		`verb=patch `+
+			`"responseStatus.code"=200 `+
+			`"objectRef.apiGroup"="appstudio.redhat.com" `+
+			`"objectRef.resource"="releases" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath path=responseObject.status.conditions{} output=conditions `+
+			`| mvexpand conditions `+
+			`| spath input=conditions `+
+			`| search type="Released" AND (reason="Succeeded" OR reason="Failed") `+
+			`| eval event="Release ".mvindex('responseObject.status.conditions{}.reason', 0)`,
+		[]string{"namespace", "name", "status_message", "status_reason"},
 	)
 	return q
 }
--- a/querygen/userjourney_fs.go
+++ b/querygen/userjourney_fs.go
@@ -37,12 +37,12 @@ var UserJourneyFieldSet = FieldSet{
 		srcExpr:   `replace('responseObject.metadata.annotations.build.appstudio.openshift.io/repo',"^([^?]*)(.*)?","\1")`,
 	},
 	"status_message": {
-		subObj:    "properties",
-		srcFields: []string{"responseObject.status.conditions{}.message"},
+		subObj:  "properties",
+		srcExpr: `mvindex('responseObject.status.conditions{}.message', 0)`,
 	},
 	"status_reason": {
-		subObj:    "properties",
-		srcFields: []string{"responseObject.status.conditions{}.reason"},
+		subObj:  "properties",
+		srcExpr: `mvindex('responseObject.status.conditions{}.reason', 0)`,
 	},
 	"target_branch": {
 		subObj:    "properties",
--- a/cmd/querygen/main.go
+++ b/cmd/querygen/main.go
@@ -62,5 +62,9 @@ func main() {
 			Title: "Build PipelineRun Completed or Failed event",
 			Query: querygen.GenBuildPipelineRunCompletedQuery(*index),
 		},
+		{
+			Title: "Release Succeeded or Failed events",
+			Query: querygen.GenReleaseCompletedQuery(*index),
+		},
 	}))
 }
--- a/querygen/querygen_test.go
+++ b/querygen/querygen_test.go
@@ -34,3 +34,8 @@ func TestGenBuildPipelineRunCompletedQuery(t *testing.T) {
 	out := GenBuildPipelineRunCompletedQuery("some_index")
 	assert.NotEqual(t, "", out)
 }
+
+func TestGenReleaseCompletedQuery(t *testing.T) {
+	out := GenReleaseCompletedQuery("some_index")
+	assert.NotEqual(t, "", out)
+}


hmariset (approved):
LGTM

querygen/querygen.go
```diff
@@ -87,3 +87,25 @@ func GenClairScanCompletedQuery(index string) string {
...
+			`"objectRef.apiGroup"="appstudio.redhat.com" `+
+			`"objectRef.resource"="releases" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath responseObject.status.conditions{} `+
+			`| mvexpand responseObject.status.conditions{} `+
+			`| search responseObject.status.conditions{}.type="Released" AND `+
+			`(responseObject.status.conditions{}.reason="Succeeded" OR responseObject.status.conditions{}.reason="Failed") `+
+			`| eval event="Release completed events"`,
```
amisstea, line 107: On second thought, I think we should make this dynamic using `responseObject.status.conditions{}.reason`. Let's see what becomes of this thread on the same topic: https://github.com/redhat-appstudio/segment-bridge/pull/34/files#r1312366672.

Also, ` events` should be omitted.

querygen/userjourney_fs.go
```diff
@@ -43,6 +43,18 @@ var UserJourneyFieldSet = FieldSet{
...
+		srcFields: []string{"responseObject.status.conditions{}.message"},
+	},
+	"release_status_reason": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.reason"},
+	},
+	"release_status_type": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.type"},
+	},
```
amisstea, line 57: We shouldn't need status type.

querygen/userjourney_fs.go
```diff
@@ -43,6 +43,18 @@ var UserJourneyFieldSet = FieldSet{
...
 		srcFields: []string{"clair_scan_result.vulnerabilities.low"},
 	},
+	"release_status_message": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.message"},
+	},
+	"release_status_reason": {
+		subObj:    "properties",
+		srcFields: []string{"responseObject.status.conditions{}.reason"},
+	},
```
amisstea, line 53: Let's rename these to `status_message` and `status_reason` to be consistent with #34 and less Release-specific.

amisstea (changes requested):
When running this...

```
./scripts/fetch-uj-records.sh | jq '.result.properties | fromjson'
```

I get results like...

```
{
  "apiGroup": "appstudio.redhat.com",
  "apiVersion": "v1alpha1",
  "application": null,
  "component": null,
  "kind": "releases",
  "release_status_message": [
    "Release processing failed",
    "",
    "Tasks Completed: 5 (Failed: 1, Cancelled 0), Skipped: 2"
  ],
  "release_status_reason": [
    "Failed",
    "Succeeded",
    "Failed"
  ],
  "release_status_type": [
    "Released",
    "Validated",
    "Processed"
  ]
}
```

We should ideally get a key/value pair for the `reason` and `message` of only the `Succeeded` condition and not these lists. Also, AFAIK, Segment doesn't support non-scalar values for properties anyway.

querygen/querygen.go
```diff
@@ -87,3 +87,25 @@ func GenClairScanCompletedQuery(index string) string {
...
+		[]string{"namespace", "application", "component", "release_status_message", "release_status_reason", "release_status_type"},
```
amisstea, line 108: `application` and `component` are always null. It's probably because the `Release` CRs don't have the respective annotations. Please look for any useful annotations in their place or omit them here.

querygen/querygen.go
```diff
@@ -87,3 +87,25 @@ func GenClairScanCompletedQuery(index string) string {
...
+			`"objectRef.apiGroup"="appstudio.redhat.com" `+
+			`"objectRef.resource"="releases" `+
+			`"objectRef.subresource"="status" `+
+			`"responseObject.metadata.resourceVersion"="*" `+
+			`"responseObject.status.completionTime"="*" `+
+			`| spath responseObject.status.conditions{} `+
+			`| mvexpand responseObject.status.conditions{} `+
+			`| search responseObject.status.conditions{}.type="Released" AND `+
+			`(responseObject.status.conditions{}.reason="Succeeded" OR responseObject.status.conditions{}.reason="Failed") `+
+			`| eval event="Release completed ".mvindex('responseObject.status.conditions{}.reason', 0)`,
```
amisstea, line 107: This will result in events named `Release completed Failed` and `Release completed Succeeded` which is confusing. Let's drop the `completed`.

querygen/querygen.go
```diff
@@ -87,3 +87,25 @@ func GenClairScanCompletedQuery(index string) string {
...
+		[]string{"namespace", "status_message", "status_reason"},
```
amisstea, line 108: Let's also include `"name"` in the list of output fields. Otherwise the resulting events all look practically the same.

Kousalya1998:
> When running this...
> 
> ```
> ./scripts/fetch-uj-records.sh | jq '.result.properties | fromjson'
> ```
> 
> I get results like...
> 
> ```
> {
>   "apiGroup": "appstudio.redhat.com",
>   "apiVersion": "v1alpha1",
>   "application": null,
>   "component": null,
>   "kind": "releases",
>   "release_status_message": [
>     "Release processing failed",
>     "",
>     "Tasks Completed: 5 (Failed: 1, Cancelled 0), Skipped: 2"
>   ],
>   "release_status_reason": [
>     "Failed",
>     "Succeeded",
>     "Failed"
>   ],
>   "release_status_type": [
>     "Released",
>     "Validated",
>     "Processed"
>   ]
> }
> ```
> 
> We should ideally get a key/value pair for the `reason` and `message` of only the `Succeeded` condition and not these lists. Also, AFAIK, Segment doesn't support non-scalar values for properties anyway.

@amisstea The ticket says we need to produce segment events when a Release succeeds or fails, so I have an OR condition that basically filters out both Succeeded and Failed reason results. Do I need to update the command to just Succeeded?

amisstea:
>> We should ideally get a key/value pair for the reason and message of only the Succeeded condition and not these lists. Also, AFAIK, Segment doesn't support non-scalar values for properties anyway.

> @amisstea The ticket says we need to produce a segment event when a Release succeeds or fails, so I have an OR condition that basically filters out both Succeeded and Failed reason results. Do I need to update the command to just Succeeded?

Oh, sorry, that's my mistake. I meant to say that we should get a single value for `status_reason` and `status_message` from the condition with the `Released` type. We still want to produce events when `reason` is either `Succeeded` or `Failed`.

querygen/userjourney_fs.go
ShakedAviv50313, line 1: "status_message" and "status_reason" now exist because RHTAPWATCH-151 was merged

ShakedAviv50313 (commented):
LGTM,
I'll approve this PR once "status_message" and "status_reason" are removed, and the issue with the 3 values is fixed.

Kousalya1998:
@amisstea I have now updated the query as we discussed over the call. PTAL and let me know your thoughts.

amisstea:
@Kousalya1998 there's still the issue with getting multiple values in the output for `status_reason` and `status_message`.

One _imperfect_ solution (until https://issues.redhat.com/browse/RHTAPWATCH-494 is resolved) is to use a `srcExpr` for these fields and assume we only care about the first value. Ex:
```
	"status_reason": {
		subObj: "properties",
		srcExpr: `mvindex('responseObject.status.conditions{}.reason', 0)`,
	},
```

Wdyt? Have you tried to solve this yet?

Kousalya1998:
> @Kousalya1998 there's still the issue with getting multiple values in the output for `status_reason` and `status_message`.
> 
> One _imperfect_ solution (until https://issues.redhat.com/browse/RHTAPWATCH-494 is resolved) is to use a `srcExpr` for these fields and assume we only care about the first value. Ex:
> 
> ```
> 	"status_reason": {
> 		subObj: "properties",
> 		srcExpr: `mvindex('responseObject.status.conditions{}.reason', 0)`,
> 	},
> ```
> 
> Wdyt? Have you tried to solve this yet?

@amisstea Will check this one now.

Kousalya1998:
@amisstea @ShakedAviv50313 @hmariset Need your reviews!!

querygen/userjourney_fs.go
```diff
@@ -41,8 +41,8 @@ var UserJourneyFieldSet = FieldSet{
 		srcFields: []string{"responseObject.status.conditions{}.message"},
 	},
 	"status_reason": {
-		subObj:    "properties",
-		srcFields: []string{"responseObject.status.conditions{}.reason"},
+		subObj:  "properties",
+		srcExpr: `mvindex('responseObject.status.conditions{}.reason', 0)`,
```
amisstea, line 45: We need to use the same `mvindex` workaround on the `status_message` FieldSet entry too.

amisstea (changes requested):

querygen/userjourney_fs.go
Kousalya1998: I got something like and didn't notice it:

```
{
  "apiGroup": "appstudio.redhat.com",
  "apiVersion": "v1alpha1",
  "kind": "releases",
  "name": "fbc-stage-index-manual-release",
  "status_message": "",
  "status_reason": "Succeeded"
}
```

amisstea (approved):

PR #53, opened 2023-09-19
fix(RHTAPWATCH-522): container builds flakiness

Omeramsc:
# Description

The Splunk container had a few issues that caused flakiness:

- The second stage of the dockerfile did not always finished copying
the DB (digested logs) before finishing the build process, leading
to failures.
- `podman kube play` also seems to cause instability when building
the image.

This change solve those issues by:
1. Add a 10 second wait time between copying the Splunk DB
and finishing the build process; for now, it seems to be enough
time for the process to complete consecutively.
2. Change the test framework so instead of relaying on `podman kube play`
for both building and running the container, the framework will build
using `podman build`, and run the container with `podman kube play`
using the built image.

[RHTAPWATCH-522](https://issues.redhat.com/browse/RHTAPWATCH-522)

## Type of change

# How Has This Been Tested?

- Manually building the container using `podman play kube`
- Manually building the container using `podman build` for the build process, and `podman play kube` for running the container.
- Running the container using the automated splunk tests

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added meaningful description with JIRA/GitHub issue key (if applicable)

### DIFF
--- a/containerfixture/containerfixture.go
+++ b/containerfixture/containerfixture.go
@@ -57,9 +57,23 @@ func generateFixturePodYaml(templateYaml string) FixtureInfo {
 	return deployment
 }
 
+// Building a container image using the 'podman build' command.
+func buildContainer(t *testing.T, serviceName, service_image_dir string) {
+	cmd := exec.Command("podman", "build", "-t", serviceName, service_image_dir)
+	stdout, err := cmd.Output()
+	if err == nil {
+		t.Logf("STDOUT: %s", string(stdout))
+		return
+	}
+	t.Errorf("Failed to build image")
+	t.Errorf("STDERR: %s", err.Error())
+	t.Fail()
+}
+
 // Building and Running a k8s manifest using 'podman kube play' command.
+// if the image is already built, it'll run the pod using it.
 // It will make up to three attempts to run with different configurations,
-// allowing it to run concurrently if necessary
+// allowing it to run concurrently if necessary.
 func buildAndRunPod(t *testing.T, manifestTemplate string) FixtureInfo {
 	var deployment FixtureInfo
 	for attempt := 0; attempt < containerBuildAttempts; attempt++ {
@@ -130,7 +144,8 @@ func cleanup(t *testing.T, podName string) {
 	}
 }
 
-func WithServiceContainer(t *testing.T, ServiceManifest string, testFunc func(FixtureInfo)) {
+func WithServiceContainer(t *testing.T, serviceName, serviceDirPath, ServiceManifest string, testFunc func(FixtureInfo)) {
+	buildContainer(t, serviceName, serviceDirPath)
 	deployment := buildAndRunPod(t, ServiceManifest)
 
 	defer cleanup(t, deployment.PodName)
--- a/splunk/Dockerfile
+++ b/splunk/Dockerfile
@@ -9,7 +9,7 @@ COPY --chmod=777 tests/indexes.conf /opt/splunk/etc/apps/search/local/indexes.co
 COPY --chmod=777 scripts/log_indexing.sh /opt/splunk/log_indexing.sh
 COPY --chmod=777 scripts/pre-run.sh /opt/splunk/pre-run.sh
 
-RUN microdnf install -y jq \
+RUN microdnf install -y jq --disablerepo="*" --enablerepo="ubi-8-appstream-rpms" \
     && microdnf clean all \
     && umask 000 \
     && touch "/opt/splunk/output.log"
@@ -28,7 +28,7 @@ USER root
 
 RUN mkdir -p /opt/splunk/var/lib/splunk
 COPY --chmod=777 --from=builder /opt/splunk/etc/apps/search/local/indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
-COPY --chmod=777 --from=builder /var/splunk_buildtime_db/splunk /opt/splunk/var/lib/splunk/
+COPY --chmod=777 --from=builder /var/splunk_buildtime_db/splunk/ /opt/splunk/var/lib/splunk/
 
 ENV SPLUNKD_SSL_ENABLE=false
 
--- a/splunk/scripts/pre-run.sh
+++ b/splunk/scripts/pre-run.sh
@@ -21,9 +21,12 @@ done
 # Index the logs required for the tests
 bash /opt/splunk/log_indexing.sh
 
+# Wait an extra 10 seconds to assure completion of logs ingestion (fix flakiness)
+sleep 10
+
 # Copy the whole Splunk DB to an accessible directory
-sudo mkdir -p /var/splunk_buildtime_db/splunk
-sudo cp -r /opt/splunk/var/lib/splunk/ /var/splunk_buildtime_db/splunk
+sudo mkdir -p /var/splunk_buildtime_db
+sudo cp -r /opt/splunk/var/lib/splunk/ /var/splunk_buildtime_db
 
 # Shut down the Splunk service
 sudo -u splunk /opt/splunk/bin/splunk stop
--- a/kwok/getuidmap_test.go
+++ b/kwok/getuidmap_test.go
@@ -5,35 +5,41 @@ import (
 	"testing"
 
 	"github.com/redhat-appstudio/segment-bridge.git/containerfixture"
+	"github.com/redhat-appstudio/segment-bridge.git/utils"
 	"github.com/stretchr/testify/assert"
 )
 
 //go:embed kwok_container_template.yml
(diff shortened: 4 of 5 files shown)


containerfixture/containerfixture.go
```diff
@@ -130,7 +144,8 @@ func cleanup(t *testing.T, podName string) {
 	}
 }
 
-func WithServiceContainer(t *testing.T, ServiceManifest string, testFunc func(FixtureInfo)) {
+func WithServiceContainer(t *testing.T, serviceName, serviceDirPath, ServiceManifest string, testFunc func(FixtureInfo)) {
+	buildContainer(t, serviceName, serviceDirPath)
```
ifireball, line 148: The one benefit of `kube play` is that it can skip building if its not necessary, can this do that too?

containerfixture/containerfixture.go
```diff
@@ -130,7 +144,8 @@ func cleanup(t *testing.T, podName string) {
 	}
 }
 
-func WithServiceContainer(t *testing.T, ServiceManifest string, testFunc func(FixtureInfo)) {
+func WithServiceContainer(t *testing.T, serviceName, serviceDirPath, ServiceManifest string, testFunc func(FixtureInfo)) {
```
ifireball, line 147: Can we be a little bit smarter and avoid changing the function signature here?

ifireball (changes requested):

splunk/scripts/pre-run.sh
```diff
@@ -21,9 +21,12 @@ done
 # Index the logs required for the tests
 bash /opt/splunk/log_indexing.sh
 
+# Wait an extra 10 seconds to assure completion of logs ingestion (fix flakiness)
+sleep 10
```
ifireball, line 25: did we have a loop somewhere to check if the indexing was done in order to resolve this?

containerfixture/containerfixture.go
Omeramsc: Do we really want to add another wrapper on top of that? We still need a service container, but we need to build it first now.

splunk/scripts/pre-run.sh
Omeramsc: indexing != copying and ingesting the DB files
Currently this is the most suitable solution I found.

containerfixture/containerfixture.go
ifireball: I don't understand how did my comment imply we need another wrapper

splunk/scripts/pre-run.sh
ifireball: Its not a good solution  - you just introduced a 10 second hard-set delay to all CI jobs ever.

How do YOU know that the ingestion is done?

Does Splunk perhaps has a clean shut-down option that ensures indexing and ingestion is done before it shuts down?

Omeramsc: While I get why it is not the best in terms of CI - all the other checks, that wait for Splunk to give a response - loop with a delay of 5 seconds on each iteration. this is also not the best but it's something we can work with. 

Since what we are doing is a kind of a "hack", I didn't find a  good way to verify it. maybe only looping (again) to see if `stats count` return anything other than 0 (as it seems like it either ingest all or nothing, this can be an indicator). 

as for the last suggestion, no, I did not find something like this, and due to the nature of this change I'm not sure it'll work even if it does exist.

Omeramsc: Actually - Speaking of loops, I think the bug resides in the log indexing verification process - it seems like it broke the loop when the check returned False, instead of True. No need for extra wait time,  fixed.

ifireball: See , you just saved your future self a lot of seconds...

containerfixture/containerfixture.go
Omeramsc: either way we'll need the path to the images, therefor changing the signature. unless  we decide to make assumptions (= the path to the dockerfile is always `../<service_name>`)

ifireball: you can figure out the path to the image from the YAML like Podman does.

Its `./image_name` iff `./image_name/Dockerfile` or `./image_name/Containerfile` exists (We don't really have to support `Containerfile` ATM though).

containerfixture/containerfixture.go
ifireball: did you check wither the fact that we have already built all the image layers means that the build is by-and-large skipped automatically?

Omeramsc: the build process will be mostly reusing the existing layers,  and it won't rebuild them. so the whole build itself won't be skipped, but it'll be very fast.

ifireball: did you see that in action or are you just assuming it would happen?

Omeramsc: I'm building the images and running the containers over and over again; Yes, I saw that in action.

containerfixture/containerfixture.go
Omeramsc: That's the thing - the WD is not the root directory of the repo, but the directory the test resides in (running `go test ./...` will have `./splunk` as the WD when running the Splunk test, and `./kwok` when running the kwok test)
it might actually make it easier, meaning we can just build from `./Dockerfile`, but it won't necessarily work with multiple images, or future tests that may need multiple builds from different directory. 
however, we can get the caller path (`containerfixture`), and do what you suggested from the parent directory (the root dir)

ifireball: How did you manage to make `kube play` work with it? You didn't need to pass anyhthing of that sort whe you were using it.

ifireball: It may simplify things to set PWD to a fixed location when running kube play - you can figure out where to set it to via a similar trick to what I did for `scripts/runner` and then set it in the `Cmd.Dir` field whin invoking `podman`.

Omeramsc: - we never ran multiple images from kube play before, so it ran the `splunk` image from current directory, since it was root/splunk anyway. 
- I'm not sure I understand what in `scripts/runner` is relevant? I do intend to use `Cmd.Dir` on the calls to podman, but my current method of getting the root directory (best for podman fixed location IMO) is using `filepath.Dir` twice on `runtime.Caller(0)` (`<path_to_root>/containerfixture/containerfixture.go` -> `<path_to_root>`)
- [Parsing a YAML](https://zetcode.com/golang/yaml/) file isn't that straightforward and clean in GO; I don't mind implementing it, but since I'm off for the next week and this PR is a blocker for both Sri and Homaja, I offer getting the service name as an argument for now (as we don't have multiple-services yamls currently), will open a ticket and fix it once I'm back with a better solution.

ifireball: I'm not convinced this PR is a blocker. @srivickynesh and @hmariset can unblock themselves by manually pre-building the images before running the test.

Omeramsc: Modified the solution to parse the yaml and take out the image names.

amisstea (approved):
LGTM

splunk/Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY --chmod=777 tests/indexes.conf /opt/splunk/etc/apps/search/local/indexes.co
 COPY --chmod=777 scripts/log_indexing.sh /opt/splunk/log_indexing.sh
 COPY --chmod=777 scripts/pre-run.sh /opt/splunk/pre-run.sh
 
-RUN microdnf install -y jq \
+RUN microdnf install -y jq --disablerepo="*" --enablerepo="ubi-8-appstream-rpms" \
```
srivickynesh, line 12: why do we need to disable this repo? Does this get picked from here?

srivickynesh (approved):
LGTM

containerfixture/containerfixture.go
```diff
@@ -131,10 +150,41 @@ func cleanup(t *testing.T, podName string) {
...
+	}
 	deployment := buildAndRunPod(t, ServiceManifest)
 
 	defer cleanup(t, deployment.PodName)
-
-	testFunc(deployment)
 	return
 }
+
+type PodDefinition struct {
```
ifireball, line 169: @Omeramsc I'm not convinced these annotations are needed since AFAIK the YAML and JSON readers know how to assume the field names are lower case in the data. Did you try running the code without them?

Omeramsc: It does seem to work without the annotation, I understood differently from the [documentation](https://pkg.go.dev/gopkg.in/yaml.v2#Unmarshal) though. 
anyway, fixed that.

ifireball (approved):

PR #61, opened 2023-09-27
chore(RHTAPWATCH-132): Fit to RH policy

ifireball:
Make our main image match the requirements of the Red Hat Enterprise Policy so that it can be released via RHTAP.


yftacherzog (approved):

Omeramsc (approved):
LGTM

ifireball:
Merging this even though is does not fix all the policy issues, because it does fix some of them.

PR #64, opened 2023-10-02
chore(RHTAPWATCH-386): events for PR originated build PipelineRuns

ShakedAviv50313:
Producing a segment event whenever a build PipelineRun is originated from a PR, which means it was triggered by a webhook or by a change to a PR.

### DIFF
--- a/querygen/querygen.go
+++ b/querygen/querygen.go
@@ -34,7 +34,8 @@ func GenBuildPipelineRunCreatedQuery(index string) string {
 				`"responseObject.metadata.resourceVersion"="*"`,
 		).
 		WithEventExpr(`"Build PipelineRun created"`).
-		WithFields("namespace", "application", "component", "repo", "commit_sha", "target_branch").
+		WithFields("namespace", "application", "component", "repo", "commit_sha", "target_branch",
+			"git_trigger_event_type", "git_trigger_provider", "pipeline_log_url").
 		String()
 	return q
 }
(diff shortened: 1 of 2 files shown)


amisstea (changes requested):

amisstea (changes requested):

querygen/querygen.go
```diff
@@ -34,7 +34,8 @@ func GenBuildPipelineRunCreatedQuery(index string) string {
 				`"responseObject.metadata.resourceVersion"="*"`,
 		).
 		WithEventExpr(`"Build PipelineRun created"`).
-		WithFields("namespace", "application", "component", "repo", "commit_sha", "target_branch").
+		WithFields("namespace", "application", "component", "repo", "commit_sha", "target_branch",
+			"git_trigger_event_type", "git_trigger_provider").
```
amisstea, line 38: I'm thinking we should add these fields to the other build PipelineRun related queries as well. They might be useful somehow.

amisstea (changes requested):
While it's nice to know that a build PipelineRun was created by a push or pull request event from github, I think we should provide some more detail about the git event. There are a lot more `pipelinesascode.tekton.dev/*` annotations that could be included in the properties. Please take a look and suggest some that might help the end user navigate to the origin.

querygen/querygen.go
ShakedAviv50313: Now it's in all of the Build piplineRun's related queries,
I also added pipeline log-url, other than that i'm not sure what is there more to add (there's already the commit sha, branch, repo, target branch, etc)

ifireball (approved):

PR #65, opened 2023-10-02
chore(RHTAPWATCH-132): Use OSP image from Quay

ifireball:
That image may have a better chance of passing the Clair scans and therefore the Red Hat EC checks.


Omeramsc:
Is it passing the clair-scan? what does passing this scan depends on?

ifireball:
> Is it passing the clair-scan? what does passing this scan depends on?

no

That depends on:
1. Having a base image that has an A rating in the RH catalog (Which means it passes the RH Clair scan)
2. Having the RHTAP scan be fixed to match what the RH scan does so it does not fail on false positives.

PR #67, opened 2023-10-09
Update Go Toolset image

ifireball:
The one we used seems to be blacklisted in some way, also switch to the UBI based image so we can get it from the free registry.


ifireball:
Merging to enable builds to pass

PR #69, opened 2023-10-12
chore(RHTAPWATCH-508): run yamllint checks for opened PR

Omeramsc:
# Description

Creates a github workflow which run yamllint checks on every opened pull request.

[RHTAPWATCH-508](https://issues.redhat.com/browse/RHTAPWATCH-508)

## Type of change

# Checklist:


.github/workflows/yamllint.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+      - name: Install yamllint
+        run: pip install yamllint
```
ifireball, line 18: Is there no Go equivalent we can use?

If we're gonna be using Python-based tools for a Golang repo, migt as well use [pre-commit][https://pre-commit.com/] and enjoy its extensive ecosystem and the ability to also run locally.

Omeramsc: I didn't find anything Go-equivalent, almost anything is ending up using the same yamllint tool. we might use [this](https://golangci-lint.run/) to lint go itself.

There are a lot of python-based tools we can use (like pre-commit you mentioned), I'm not convinced it is good to limit ourselves to golang tools only because the repo itself uses golang.

ifireball: Adding a Python-based tool turns Python (the interpreter) into a dependency, and also forces us to consider how to manage the tools and packages we want to use.

So again, if we must, lets use `pre-commit` which has a rather elegant solution for the problem of installing Python tools and also allows us to run all the tests locally as well as in CI.

Omeramsc: Will do
as for Yaml checks, I see we have hundreds of errors, mostly in the tekton/deployment yamls, is there any automatic formatter for yamls we wish to use, or will we ignore those?

Omeramsc: most of them are indentation and long lines,
fixing them manually will take hours, that's why I'd prefer a formatter.

ifireball: I don't know of one, maybe you can make VScode help you

yftacherzog:
@Omeramsc if I understand correctly, up until now, we were using tekton pipelines. What's the reasoning behind doing this with github workflow? I think there's simplicity in sticking to one method.

Omeramsc:
> @Omeramsc if I understand correctly, up until now, we were using tekton pipelines. What's the reasoning behind doing this with github workflow? I think there's simplicity in sticking to one method.

Currently we [cannot add custom checks to the RHTAP pipeline](https://issues.redhat.com/browse/RHTAP-1707) we are using for the other tests. This is a fast, viable solution to provide immediate value;  we might change to the RHTAP pipeline once it will be viable.

.github/workflows/yamllint.yml
Omeramsc: formatted with YAML sort and some manual changes.
also added a rule that allow spaces in braces (which we are using in templates)

.github/workflows/precommit.yaml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Install Python dependencies
+        run: pip install -r requirements.txt
```
ifireball, line 21: did you check if there is a container we can use with Python and pre-commit already installed, instead of installing them on the ubuntu machine for each run?

requirements.txt
```diff
@@ -0,0 +1 @@
+pre-commit
```
ifireball, line 1: Use full version numbers here so we get the same version for each run. If we want to use requirements.txt rather then Pipfile (pipenv) of pyproject.toml (poetry), then we probably should use pip-tools to get the usual unversioned+lock files combination.

splunk/splunk_container_template.yaml
```diff
@@ -2,21 +2,20 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: {{ .PodName}}
+  name: { { .PodName } }
```
ifireball, line 5: these changes will make this file no longer function as a proper go template. Perhaps we should change the extension for this file because its not a legit YAML file.

ifireball (changes requested):

.github/workflows/precommit.yaml
Omeramsc: I didn't find any official container recommended for this use, only random docker-hub ones. I did see a lot of repositories using the same method of installing on a fresh ubuntu machine.

requirements.txt
Omeramsc: fixed, chose to use `.tmpl.`

splunk/splunk_container_template.yaml
Omeramsc: added `requirements.lock` with a specific version, changed the workflow to use it instead.
I tried to use pip-tools to compile the requirements file, but currently encountered conflicts from pre-commit's requirements file and `importlib-resources<5.3,>=5.4`.

ifireball (approved):

ifireball:
not sure why the RHTAP pipeline is not running for this, try rebasing to see if that sorts this out.

Omeramsc:
@ifireball no Idea, also tried to rebase / retest/ push new changes

Omeramsc:
@ifireball  I reverted the formatting for `.tekton` (and set `yamllint` to ignore it completely) - the CI now works. seems like that was the issue

yftacherzog:
@Omeramsc I imagine that the new action is not running due to permission issues. Maybe you can create a PR towards your fork and verify there that it's running as expected.

yftacherzog (approved):

PR #70, opened 2023-10-16
Add the `workspaceID` property

ifireball:
This will allow grouping events by user workspaces and not just by individual users.

### DIFF
--- a/querygen/ujquery.go
+++ b/querygen/ujquery.go
@@ -37,6 +37,7 @@ func NewUserJourneyQuery(index string) *UserJourneyQuery {
 			"event_verb",
 			"kind",
 			"messageId",
+			"namespace",
 			"timestamp",
 			"type",
 			"userAgent",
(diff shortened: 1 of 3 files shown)


querygen/ujquery.go
```diff
@@ -37,6 +37,7 @@ func NewUserJourneyQuery(index string) *UserJourneyQuery {
 			"event_verb",
 			"kind",
 			"messageId",
+			"namespace",
```
Omeramsc, line 40: not sure I'm following - the `namespace` field was removed from multiple functions, why is it here? should'nt it include workspaceID?

ifireball: There is no workspaceID field in Splunk, so we query the namespace and convert it to workspaceID via jq.

I removed it elsewhere because here it adds it to all queries

Omeramsc (approved):
LGTM

PR #71, opened 2023-10-17
 chore(RHTAPWATCH-177): run gitlint check for opened PR

Omeramsc:
# Description

Adds a step to the pre-commit github workflow to run gitlint check on every opened pull request's commit message.

depends on #69 

[RHTAPWATCH-177](https://issues.redhat.com/browse/RHTAPWATCH-177)

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?
Running pre-commit and gitlint manually.

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/.github/workflows/precommit.yaml
+++ b/.github/workflows/precommit.yaml
@@ -21,3 +21,7 @@ jobs:
         run: pip install -r requirements.lock
       - name: Run pre-commit
         run: pre-commit run --all-files
+      - name: Run gitlint on the latest commit message
+        run: >-
+          pre-commit run gitlint --hook-stage commit-msg --commit-msg-filename
+          .git/COMMIT_EDITMSG
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,10 @@ repos:
       - id: yamllint
     repo: https://github.com/adrienverge/yamllint
     rev: v1.29.0
+  - hooks:
+      - id: gitlint
+        name: gitlint
+        entry: gitlint
+        stages: [commit-msg]
+    repo: https://github.com/jorisroovers/gitlint
+    rev: v0.18.0
--- a/.tekton/segment-bridge-pull-request.yaml
+++ b/.tekton/segment-bridge-pull-request.yaml
@@ -1,14 +1,16 @@
+---
 apiVersion: tekton.dev/v1beta1
 kind: PipelineRun
 metadata:
   annotations:
-    build.appstudio.openshift.io/repo: https://github.com/redhat-appstudio/segment-bridge?rev={{revision}}
-    build.appstudio.redhat.com/commit_sha: '{{revision}}'
-    build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
-    build.appstudio.redhat.com/target_branch: '{{target_branch}}'
+    build.appstudio.openshift.io/repo: >-
+      https://github.com/redhat-appstudio/segment-bridge?rev={{revision}}
+    build.appstudio.redhat.com/commit_sha: { { revision } }
+    build.appstudio.redhat.com/pull_request_number: { { pull_request_number } }
+    build.appstudio.redhat.com/target_branch: { { target_branch } }
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-event: '[pull_request]'
-    pipelinesascode.tekton.dev/on-target-branch: '[main]'
+    pipelinesascode.tekton.dev/on-event: "[pull_request]"
+    pipelinesascode.tekton.dev/on-target-branch: "[main]"
   creationTimestamp: null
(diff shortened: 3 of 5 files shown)


.github/workflows/precommit.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Install Python dependencies
+        run: pip install -r requirements.txt
```
ifireball, line 21: did you check if there is a container we can use with Python and pre-commit already installed, instead of installing them on the ubuntu machine for each run?

splunk/splunk_container_template.yaml
```diff
@@ -2,21 +2,20 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: {{ .PodName}}
+  name: { { .PodName } }
```
ifireball, line 5: these changes will make this file no longer function as a proper go template. Perhaps we should change the extension for this file because its not a legit YAML file.

requirements.txt
```diff
@@ -0,0 +1 @@
+pre-commit
```
ifireball, line 1: Use full version numbers here so we get the same version for each run. If we want to use `requirements.txt` rather then `Pipfile` (pipenv) of `pyproject.toml` (poetry), then we probably should use pip-tools to get the usual un-versioned+lock files combination.

ifireball (changes requested):

.github/workflows/precommit.yaml
Omeramsc: I didn't find any official container recommended for this use, only random docker-hub ones. I did see a lot of repositories using the same method of installing on a fresh ubuntu machine.

splunk/splunk_container_template.yaml
ifireball: `.templ` or `.got` I suppose - unless you manage to find some other convention for naming Go template files.

ifireball: With Jinja2 to convention was to use `.j2` so for example `foo.bar.j2` would be a template for generating `foo.bar`

Omeramsc: fixed, chose to use `.tmpl`.

requirements.txt
Omeramsc: added `requirements.lock` with a specific version, changed the workflow to use it instead. 
I tried to use `pip-tools` to compile the requirements file, but currently encountered conflicts from pre-commit's requirements file and `importlib-resources<5.3,>=5.4`.

Omeramsc:
@ifireball changes mentioned above are maid in #69 , will be rebased on top of it.

.github/workflows/precommit.yaml
```diff
@@ -21,3 +21,7 @@ jobs:
         run: pip install -r requirements.txt
       - name: Run pre-commit
         run: pre-commit run --all-files
+      - name: Run pre-commit on the latest commit message
+        run:
+          pre-commit run gitlint --hook-stage commit-msg --commit-msg-filename
```
yftacherzog, line 26: Are we using the default gitlint rules here? Do they fit us?

.github/workflows/precommit.yaml
```diff
@@ -21,3 +21,7 @@ jobs:
...
+          .git/COMMIT_EDITMSG
```
yftacherzog, line 27: Isn't this part of the previous line? Does it work this way?

Omeramsc: You're right, I forgot to set it as a multi-line. fixed.

.github/workflows/precommit.yaml
Omeramsc: Currently using default ones, yes; they should fit our contribution guide.

ifireball (changes requested):
why are we still seeing the YAML changes here?

Omeramsc:
@ifireball  Something is very wrong with rebasing this PR, tried multiple times but got problematic results on each. I'll be closing this for the sake of a fixed PR.

yftacherzog:
> @ifireball Something is very wrong with rebasing this PR, tried multiple times but got problematic results on each. I'll be closing this for the sake of a fixed PR.

You don't have to close the PR to fix something like this. Whatever you pushed to the branch of the new PR you could have force-pushed into this branch and the result would have been the same.

Omeramsc:
> > @ifireball Something is very wrong with rebasing this PR, tried multiple times but got problematic results on each. I'll be closing this for the sake of a fixed PR.
> 
> You don't have to close the PR to fix something like this. Whatever you pushed to the branch of the new PR you could have force-pushed into this branch and the result would have been the same.

Noted

PR #74, opened 2023-10-23
refator(RHTAPWATCH-547): Per-K8s API field sets

ifireball:
Introduce a new K8sAuditFieldSet object built on top of our exiting FieldSet object, but specialized for queries against the audit log. A key feature is the ability to specify different field settings for different K8s resource types. This will make it easier to have common output fields that require different ways to calculate them for different resources.

As part of the work here, introduced the new object type and re-factored UserJourneyQuery to use it.

### DIFF
--- a/querygen/k8s_audit_fieldset.go
+++ b/querygen/k8s_audit_fieldset.go
@@ -1,0 +1,45 @@
+package querygen
+
+import (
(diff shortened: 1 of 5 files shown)


amisstea (approved):
Looks great! :+1:

PR #80, opened 2023-10-25
chore(RHTAPWATCH-177): run gitlint on CI

Omeramsc:
# Description

Fix an issue where the pre-commit github action that runs gitlint skipped linting the commit message.

This was done by installing an additional pre-commit gitlint hook that is designed for ci specifically.

## Type of change

# How Has This Been Tested?
Running the pre-commit hook locally

# Checklist:

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -1,5 +1,5 @@
 ---
-name: Pre-commit
+name: checks on Pull Request
 "on":
   push:
     branches:
@@ -23,7 +23,18 @@ jobs:
         run: pip install -r requirements.lock
       - name: Run pre-commit
         run: pre-commit run --all-files
-      - name: Run gitlint on the latest commit message
+  gitlint:
+    name: Run gitlint checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Install Python dependencies
+        run: pip install -r requirements.lock
+      - name: Run gitlint on CI with pre-commit
         run: >-
-          pre-commit run gitlint --hook-stage commit-msg --commit-msg-filename
-          .git/COMMIT_EDITMSG
+          gitlint --commits origin/${{ github.event.pull_request.base.ref
+          }}..HEAD
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,15 @@
+[general]
+contrib=contrib-title-conventional-commits,contrib-body-requires-signed-off-by
+
+[title-max-length]
+line-length=72
+
+[body-max-line-length]
+line-length=72
+
+[body-first-line-empty]
+
+# Dependabot tends to generate lines that exceed the default 80 char limit.
+[ignore-by-author-name]
+regex=dependabot
+ignore=all
--- a/requirements.lock
+++ b/requirements.lock
@@ -1,1 +1,2 @@
 pre-commit==3.5.0
+gitlint==0.18.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 pre-commit
+gitlint


.github/workflows/precommit.yaml
```diff
@@ -21,7 +21,5 @@ jobs:
         run: pip install -r requirements.lock
       - name: Run pre-commit
         run: pre-commit run --all-files
-      - name: Run gitlint on the latest commit message
-        run: >-
-          pre-commit run gitlint --hook-stage commit-msg --commit-msg-filename
-          .git/COMMIT_EDITMSG
+      - name: Run gitlint on CI with pre-commit
+        run: pre-commit run --hook-stage manual gitlint-ci
```
yftacherzog, line 25: Which commits does it check? Latest? entire log? PR content?

Omeramsc: latest commit by default. can be configured for more, but I don't think we want that.

yftacherzog: Suppose that a PR has more than one commit (happens from time to time) then we're not covered.
Also, I see for the current PR it complains over some sort of merge commit, which is not the latest commit.

I think it should test the commits between the target branch and the PR branch. I don't know how the command should look like exactly here, but the I think you get those git references "for free" inside a github action.

yftacherzog: I'm not seeing the added value for implementing this under precommit rather than a direct CI test (action or pipeline).
I feel like we're having a harder time running gitlint the way we want, with it encapsulated as a pre-commit hook.

We already have it solved in repos under redhat-appstudio both as workflow and as pipeline. Are we gaining anything with the current approach?

Omeramsc: Implementing it under precommit will allow having only one source for gitlint configuration, e.g: the version of gitlint we are using for both local and CI tests.
However I'm not against running it as a separate workflow, I'll check if the current issue can be easily solved

.pre-commit-config.yaml
```diff
@@ -13,5 +13,7 @@ repos:
         name: gitlint
         entry: gitlint
         stages: [commit-msg]
+      - id: gitlint-ci
+        args: [--commits, "origin/$(git rev-parse --abbrev-ref HEAD)..HEAD"]
```
Omeramsc, line 17: @yftacherzog This makes it check all the commits on the branch.

ifireball (approved):

ifireball:
LGTM, but perhaps we need to reach an agreement with @yftacherzog

.pre-commit-config.yaml
yftacherzog: I would expect HEAD points to the last commit on the PR branch and I suspect `origin/$(git rev-parse --abbrev-ref HEAD)` might point to the same point as well. No? (i.e. if the PR branch is foo, then it will run `--commits origin/foo..foo`)

In any case, I don't think we should check all the commits on the branch. Suppose that there was a bad commit message in the past, then it makes us stuck forward. This test should be about what is being introduced on the current PR.

Looking at checks implemented with github workflows in our org, I see them using references that github provides. I think this might be doable here as well: https://github.com/redhat-appstudio/release-service/blob/e2fd1d6ea2ac4e913647dca93b84ba41eacc3f6f/.github/workflows/pr.yml#L123

We have the implementation we did in tekton. It might help as well:
https://github.com/redhat-appstudio/o11y/blob/main/.tekton/pull-request.yaml#L81

Omeramsc: wasn't aware of this possibility, checking

Omeramsc: I don't really like maintaining gitlint on pre-commit for local checks and on a separate job for CI, which might create conflicting or missing reports if not handled correctly, but the current solution isn't ideal either. I used your solution for now.

yftacherzog: I understand. I think the part of being able to tell what's the target branch for the changes in very clear at a PR level, but might not be as clear on a local env. I think this check is ok to run only in CI.

Omeramsc: Locally there won't be a need to check all commits, as we'll be unable to commit something if it does not pass the gitlint

yftacherzog (approved):
lgtm

PR #91, opened 2023-10-29
feat(RHTAPWATCH-547) Application fields everywhere

ifireball:
Make sure all events reported to Segment include an "application" field so that we can use it as a "group value" to group events.

### DIFF
--- a/querygen/ujquery.go
+++ b/querygen/ujquery.go
@@ -82,6 +82,9 @@ var UJFieldSet = K8sAuditFieldSet{
 			srcFields: []string{"build_status.pac.merge-url"},
 		},
 	},
+	K8sApiId{"appstudio.redhat.com","applications"}: {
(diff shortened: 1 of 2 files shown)


Omeramsc (approved):
LGTM

querygen/ujquery.go
```diff
@@ -98,6 +101,12 @@ var UJFieldSet = K8sAuditFieldSet{
 				)`,
 		},
 	},
+	K8sApiId{"appstudio.redhat.com","releases"}: {
+		"application": {
+			subObj: "properties",
+			srcExpr: `spath(_raw,"responseObject.metadata.ownerReferences{0}.name")`,
```
amisstea, line 107: Why do we need `spath`? This works and should perform better since the field is already indexed.

```
eval owner_name=mvindex('responseObject.metadata.ownerReferences{}.name', 0)
```

ifireball: done.

PR #92, opened 2023-10-29
fix: gitlint workflow runs on pull request only

Omeramsc:
# Description

Separates gitlint from the workflow for pull-request and push, as it cannot and should not run on push.

The new workflow runs it only for the pull-request process itself.

## Type of change

# Checklist:

### DIFF
--- a/.github/workflows/gitlint.yaml
+++ b/.github/workflows/gitlint.yaml
@@ -1,28 +1,12 @@
 ---
-name: checks on Pull Request
+name: gitlint checks on Pull Request
 "on":
-  push:
-    branches:
-      - main
   pull_request:
     types:
       - opened
       - reopened
       - synchronize
 jobs:
-  pre-commit:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: "3.11"
-      - name: Install Python dependencies
-        run: pip install -r requirements.lock
-      - name: Run pre-commit
-        run: pre-commit run --all-files
   gitlint:
     name: Run gitlint checks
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-name: checks on Pull Request
+name: checks on Pull Request and push
```
yftacherzog, line 2: I think we should also rename the file

ifireball (approved):

.github/workflows/pull_request.yaml
Omeramsc: done

yftacherzog (commented):
I think we should decide whether we name the workflow  based on what they are doing or based on when they are running.

Accordingly, either the gitlint file can be renamed to pull_request, or the pull_request_and_push should be named according to what we're doing there.

Omeramsc:
> I think we should decide whether we name the workflow based on what they are doing or based on when they are running.
> 
> Accordingly, either the gitlint file can be renamed to pull_request, or the pull_request_and_push should be named according to what we're doing there.

Do we have further plans for checks to run during pull request **only**? 
pull_request_and_push will include more checks in the future, it cannot really include all of them, unless we call it "code quality" or something like that.

yftacherzog (approved):
> > I think we should decide whether we name the workflow based on what they are doing or based on when they are running.
> > Accordingly, either the gitlint file can be renamed to pull_request, or the pull_request_and_push should be named according to what we're doing there.
> 
> Do we have further plans for checks to run during pull request **only**? pull_request_and_push will include more checks in the future, it cannot really include all of them, unless we call it "code quality" or something like that.

Don't know. I'll let you decide how to go about this. I was initially against the inconsistent naming convention, but it might make sense. 

lgtm

PR #94, opened 2023-10-30
chore(RHTAPWATCH-150): go source checks

ShakedAviv50313:
add go source checks for the repo

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,3 +19,20 @@ repos:
       - id: go-mod-tidy
     repo: https://github.com/dnephin/pre-commit-golang
     rev: v0.5.1
+  - hooks:
+      - id: golangci-lint-full
+        args:
+          - --disable-all
+          - --verbose
+          - --enable
+          - gofmt
+          - --enable
+          - govet
+          - --enable
+          - ineffassign
+          - --enable
+          - staticcheck
+          - --enable
+          - unused
+    repo: https://github.com/golangci/golangci-lint
+    rev: v1.55.1
--- a/querygen/k8s_audit_fieldset.go
+++ b/querygen/k8s_audit_fieldset.go
@@ -24,10 +24,10 @@ func (kfs K8sAuditFieldSet) QueryGen(
 	index string, api K8sApiId, searchExpr string, fields []string, extra ...FieldSet,
 ) (string, error) {
 	searchCmd := strings.TrimSpace(fmt.Sprintf(
-		`search index="%s" log_type=audit ` +
-		`"objectRef.apiGroup"="%s" ` +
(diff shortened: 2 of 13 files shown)


ifireball:
This is probably based on a very old version of the code - you can see the pre-commit action is not even running, please rebase on top of the latest code.

ifireball:
Did you verify that the checks can actually catch issues?
Are we sure all our go files are checked?

scripts/runner.go
```diff
@@ -46,3 +46,5 @@ func GetRepoRootDir() (string, error) {
 	dirPath := filepath.Dir(filepath.Dir(filename))
 	return dirPath, nil
 }
+
+
```
Omeramsc, line 50: are those a result of changes required by the go source checks?

.pre-commit-config.yaml
```diff
@@ -15,3 +15,9 @@ repos:
         stages: [commit-msg]
     repo: https://github.com/jorisroovers/gitlint
     rev: v0.18.0
+  - hooks:
+      - id: go-fmt-import
+      - id: go-vet
+      - id: go-lint
+    repo: https://github.com/Bahjat/pre-commit-golang
+    rev: v1.0.2
```
Omeramsc, line 23: 1.0.3 is available, any reason not to use it?

.pre-commit-config.yaml
```diff
@@ -15,3 +15,9 @@ repos:
         stages: [commit-msg]
     repo: https://github.com/jorisroovers/gitlint
     rev: v0.18.0
+  - hooks:
```
Omeramsc, line 18: seems like this repo has also `go-unit-tests` and other hooks, worth looking if we can get some value from them

scripts/runner.go
ShakedAviv50313: I'm trying to verify that the go checks in this PR are actually working in the ci, these lines will be removed once I'm sure the checks work

.pre-commit-config.yaml
ifireball: Given that our unit tests also try to spin up containers, I wouldn't try to scope that into this change

.pre-commit-config.yaml
ShakedAviv50313: It is no longer being used, now we're using golangci-lint and using it's latest version.

yftacherzog:
@ShakedAviv50313 one of the files that you modified here was also modified since yesterday. You'd need to rebase.

It might be tricky.

yftacherzog (changes requested):
@ShakedAviv50313 you mixed up the content of the commits when you rebased. Ping me and we sort it out.

Omeramsc (approved):
LGTM

ifireball (approved):

PR #96, opened 2023-10-30
chore(RHTAPWATCH-132): Build image in GH actions

ifireball:
Build the main service image with GitHub actions and Push to quay.io.

### DIFF
--- a/.github/workflows/pull_request_and_push.yaml
+++ b/.github/workflows/pull_request_and_push.yaml
@@ -23,3 +23,39 @@ jobs:
         run: pip install -r requirements.lock
       - name: Run pre-commit
         run: pre-commit run --all-files
+
+  build-main-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_and_push.yaml
```diff
@@ -23,3 +23,39 @@ jobs:
...
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Log in to Red Hat Registry
+      uses: redhat-actions/podman-login@v1
+      with:
+        registry: registry.redhat.io
+        username: ${{ secrets.REGISTRY_REDHAT_IO_USER }}
```
Omeramsc, line 38: Do we have the username configured? seems like the workflow fail due to it missing

PR #97, opened 2023-10-30
chore(RHTAPWATCH-150): add go-mod-tidy

ShakedAviv50313:
add go-mod-tidy check to the repository

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,3 +15,7 @@ repos:
         stages: [commit-msg]
     repo: https://github.com/jorisroovers/gitlint
     rev: v0.18.0
+  - hooks:
+      - id: go-mod-tidy
+    repo: https://github.com/dnephin/pre-commit-golang
+    rev: v0.5.1
--- a/.github/workflows/pull_request_and_push.yaml
+++ b/.github/workflows/pull_request_and_push.yaml
@@ -22,4 +22,4 @@ jobs:
       - name: Install Python dependencies
         run: pip install -r requirements.lock
       - name: Run pre-commit
-        run: pre-commit run --all-files
+        run: pre-commit run --verbose --all-files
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -15,3 +15,8 @@ repos:
         stages: [commit-msg]
     repo: https://github.com/jorisroovers/gitlint
     rev: v0.18.0
+  - hooks:
+      - id: go-mod-tidy
+    args: [--verbose]
```
yftacherzog, line 20: Is it providing verbose output? I think the indentation is wrong.
```suggestion
  - hooks:
      - id: go-mod-tidy
        args: [--verbose]
```

yftacherzog (changes requested):
Were you able to verify that it actually doing something? One way to go about it would be to duplicate one of the lines inside the `require` block in the `go.mod` file at the root of the repo. Running go mod tidy should eliminate the duplicate line, so the test should fail.

ShakedAviv50313:
@yftacherzog I tried that and it did fail, now when it's fixed it is not failing anymore

.pre-commit-config.yaml
```diff
@@ -15,3 +15,8 @@ repos:
         stages: [commit-msg]
     repo: https://github.com/jorisroovers/gitlint
     rev: v0.18.0
+  - hooks:
+      - id: go-mod-tidy
+        args: [--verbose]
```
yftacherzog, line 20: Does this has any effect?

ShakedAviv50313: apparently not, I removed it

yftacherzog: The important bit is whether it makes a difference when there's an error. I think you should try to fail the check again and see if it makes a difference in that case.

ShakedAviv50313: I failed the check a few times and it was the same with and without it

yftacherzog (approved):
lgtm

Omeramsc (approved):
LGTM

ifireball (approved):

PR #99, opened 2023-10-31
refactor getuidmap

srivickynesh:
moved executor and removed unused const

# Description

Please include a summary of the changes and the related issue. Please also
include relevant motivation and context. List any dependencies that are
required for this change.

## Type of change

# How Has This Been Tested?

go test ./guidmap 
ok  	github.com/redhat-appstudio/segment-bridge.git/guidmap	10.663s

# Checklist:


ifireball (changes requested):
please make sure you pass the gitlint check

ifireball (changes requested):
While I don't fully understand this change because I don't know the context this code is used in, I don't think any of it belongs in the kwok directoy.

amisstea:
These small changes look fine but I share the same confusion as @ifireball. It's bizarre to see `getuidmap.go` inside the `kwok` package and the `ExecuteAndParseScript` func is neither kwok nor getuidmap specific. It seems this package should undergo some additional refactoring.

srivickynesh:
@amisstea Could you take a look again?

guidmap/getuidmap_test.go
```diff
@@ -1,19 +1,33 @@
-package kwok
+package guid
```
amisstea, line 1: We should avoid `guid` as a package name. It could cause some confusion since [UUIDs/GUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) are commonly used.

kwokfixture/kwokfixture.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+	"os/exec"
+
+	"github.com/redhat-appstudio/segment-bridge.git/testfixture"
+)
+
+const (
+	contextName      = "kwok"
+	clusterServerURL = "http://127.0.0.1:8080"
+	scriptPath       = "../scripts/get-uid-map.sh"
```
amisstea, line 13: The `kwokfixture` package shouldn't hard code this script or else it can't be reused easily. It should be passed in by the test.

amisstea (changes requested):

testfixture/testfixture.go
```diff
@@ -0,0 +1,33 @@
+package testfixture
```
ifireball, line 1: this name is too generic, try thinking of a name that better describes what this does.

kwokfixture/kwokfixture.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func runOcCommand(args ...string) error {
+	cmd := exec.Command("oc", args...)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to run oc command: %w", err)
+	}
+	return nil
+}
+
```
ifireball, line 20: I think we could and should replace the code below with a simple fixed kubeconfig file committed into the repo.

testfixture/testfixture.go
ifireball, line 1: this name is too generic, try thinking of a name that better describes what this does.

ifireball (changes requested):

testfixture/testfixture.go
srivickynesh: testfixture supposedly has generic shared libraries for the tests that we write. Unless we want to shrink only to what it does.

kwokfixture/kwokfixture.go
srivickynesh: We already have another ticket for creating kubeconfig.

ifireball: so what is the point writing code we will throw away later?

testfixture/testfixture.go
ifireball: This is like calling a library "utils" it makes it impossible for someone to know what is in it without reading all the code

kwokfixture/kwokfixture.go
srivickynesh: This is already there I am just moving it to different file for now

srivickynesh:
I will close this for now. this refactoring can be done as part of the Jira task. Currently, this task is not associated with any Jira.

Omeramsc:
@srivickynesh I don't believe this should be closed, as the current structure of code is built upon the use of kwok for uid-map only

PR #104, opened 2023-11-02
chore(RHTAPWATCH-504\505): {uid, ws}-map-maker fail on connection errors

Omeramsc:
# Description

The uid-map-maker and ws-map-maker script ignored cases where
there was a connection failure. this change
aim to catch those failures and report them
for better visibility of the error.

[RHTAPWATCH-504](https://issues.redhat.com/browse/RHTAPWATCH-504)
[RHTAPWATCH-505](https://issues.redhat.com/browse/RHTAPWATCH-505)

## Type of change

# How Has This Been Tested?

Not fully tested yet.

# Checklist:

### DIFF
--- a/scripts/uid-map-maker-job.sh
+++ b/scripts/uid-map-maker-job.sh
@@ -11,6 +11,7 @@
 #
 set -o pipefail -o errexit -o nounset -o xtrace
 
-oc create configmap --dry-run=client -o yaml uid-map \
-  --from-file=uid-map.json=<(KUBECONFIG="$KUBECONFIG_SRC" get-uid-map.sh) \
-| oc apply -f -
+if ! oc create configmap --dry-run=client -o yaml uid-map \
+  --from-file=uid-map.json=<(KUBECONFIG="$KUBECONFIG_SRC" get-uid-map.sh) | oc apply -f -; then
+  echo "Error: Encountered an issue while applying the ConfigMap."
+fi
--- a/scripts/ws-map-maker-job.sh
+++ b/scripts/ws-map-maker-job.sh
@@ -11,6 +11,7 @@
 #
 set -o pipefail -o errexit -o nounset -o xtrace
 
-oc create configmap --dry-run=client -o yaml ws-map \
-  --from-file=ws-map.json=<(KUBECONFIG="$KUBECONFIG_SRC" get-workspace-map.sh) \
-| oc apply -f -
+if ! oc create configmap --dry-run=client -o yaml ws-map \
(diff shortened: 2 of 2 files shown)


scripts/uid-map-maker-job.sh
```diff
@@ -11,6 +11,8 @@
 #
 set -o pipefail -o errexit -o nounset -o xtrace
 
-oc create configmap --dry-run=client -o yaml uid-map \
-  --from-file=uid-map.json=<(KUBECONFIG="$KUBECONFIG_SRC" get-uid-map.sh) \
-| oc apply -f -
+if ! oc create configmap --dry-run=client -o yaml uid-map \
+  --from-file=uid-map.json=<(KUBECONFIG="$KUBECONFIG_SRC" get-uid-map.sh) | oc apply -f -; then
+  echo "Error: Encountered an issue while connecting to the cluster. ConfigMap was not created."
```
yftacherzog, line 16: Is the error being ignored at the moment only for connectivity issues? Suppose that `oc apply` fails for some other reason (e.g. the cluster rejects the configs), would the script fail on that without the current change?

I think it might be better to have the error message state that it failed applying the configmap, rather than suggesting the reason.

ifireball: @Omeramsc the underlying assumption of the code you rote seems to be we get a non-zero exit code from the entire pipeline on a connectivity issue. If that is the case then the `set -o errexit` already takes care of exiting the script and the code you wrote is completely redundant.

Did you ensure `oc` actually exits with a non-zero value on cluster connectivity issues?

ifireball (changes requested):

scripts/uid-map-maker-job.sh
Omeramsc: @ifireball the "exit 1" indeed seems redundant. `oc` does exits with a non-zero on connectivity issues, therefore`set -o errexit` exit the script as well (opposed to what is described in the mentioned tickets); This change allows us to add an error message of our own, clarifying the failure, isn't that what we aimed for? (`this makes it harder to see when there is such an issue occurring`)

ifireball: Are you sure `oc` is not already printing a message that is good enough to stderr?

My aim was to make sure the script exits on cluster connectivity issues. If we are sure this already happens reliably on all kind of issues then we don't need to make any changes.

But are we really sure about that? What happens when our service account token expires for example?

ifireball: if `-o pipefail` and `-o errexit`are not doing the trick, adding an `if` around the command will probably not do it either.

Omeramsc:
Closing as the script does exit when a connection error occur as  the code is written now. it also outputs the error.

PR #117, opened 2023-11-16
Add job for checking the Go Toolset image

ifireball:
Add a job that ensure we use the same Go Toolset image when building our production image and when running tests in CI.

### DIFF
--- a/chore-scripts/check-same-go-toolset.sh
+++ b/chore-scripts/check-same-go-toolset.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#  check-same-go-toolset.sh
+#    Check that we use the same Go Toolset image in our container build and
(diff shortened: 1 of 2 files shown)


chore-scripts/check-same-go-toolset.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#  check-same-go-toolset.sh
+#    Check that we use the same Go Toolset image in our container build and
+#    while runnig our tests in CI.
```
amisstea, line 4:
```suggestion
#    while running our tests in CI.
```

amisstea (approved):

hmariset:
LGTM

hmariset (approved):

PR #120, opened 2023-11-22
refactor(RHTAPWATCH-479): kwok image use sysv-init

ifireball:
Make our kwok image use the busybox version of sys-v init to launch and monitor processes. This will allow us to make it run multiple processes.

### DIFF
--- a/kwok/Dockerfile
+++ b/kwok/Dockerfile
@@ -3,7 +3,8 @@ FROM registry.k8s.io/kwok/cluster:v0.4.0-k8s.v1.27.3
 ARG CLUSTER_NAME=kwok
 ENV CLUSTER_NAME=${CLUSTER_NAME}
 
-COPY --chmod=755 user-signups.sh entrypoint.sh toolchain.dev.openshift.com_usersignups.yaml ./
+COPY --chmod=755 user-signups.sh toolchain.dev.openshift.com_usersignups.yaml ./
+COPY --chmod=644 inittab /etc/inittab
 
 RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
     kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
@@ -11,4 +12,4 @@ RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || 
     rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
     kwokctl stop cluster --name="$CLUSTER_NAME"
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/sbin/init"]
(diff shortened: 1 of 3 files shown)


kwok/Dockerfile
```diff
@@ -4,11 +4,12 @@ ARG CLUSTER_NAME=kwok
...
+COPY --chmod=644 inittab /etc/inittab
 
 RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
     kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
     sh ./user-signups.sh && \
     rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
     kwokctl stop cluster --name="$CLUSTER_NAME"
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/sbin/init"]
```
yftacherzog, line 15: If I understand correctly, the `ENTRYPOINT ["/entrypoint.sh"]` part is misleading. We're not running our `entrypoint.sh`, but the [one coming from the base image](https://github.com/kubernetes-sigs/kwok/blob/main/images/cluster/entrypoint.sh).

With this change, are we still starting it?

yftacherzog: (or still running something that does a similar job)

srivickynesh: yes @yftacherzog .

ifireball: no, we are overriding the one in the base image with our own during the image build. We have no need for the functionality in the base image entry point which has to do with setting up a cluster which we do during the build.

srivickynesh: Only change why it couldn't reload inittab might be on line 15. https://github.com/srivickynesh/segment-bridge/pull/1/files

yftacherzog: I don't think our `entrypoint.sh` is being copied to the image, so the existing functionality is not using those lines that this PR is putting in `kwok/inittab`.

ifireball: @srivickynesh you are not running init at all in that patch, and you've added a lot of unnecessary stuff that probably causes issues such as command for starting TTYs.

srivickynesh: those are all the defaults that comes within inittab

yftacherzog: ok. I realize now that the line of copying the entrypoint was not a part of the code I was dealing with yesterday.

kwok/Dockerfile
```diff
@@ -4,11 +4,12 @@ ARG CLUSTER_NAME=kwok
 ENV CLUSTER_NAME=${CLUSTER_NAME}
 
 COPY --chmod=755 user-signups.sh entrypoint.sh toolchain.dev.openshift.com_usersignups.yaml ./
```
srivickynesh, line 6:
```suggestion
COPY --chmod=755 user-signups.sh toolchain.dev.openshift.com_usersignups.yaml ./
```

srivickynesh: Also we need to remove kwok/entrypoint.sh as part of this change.

srivickynesh (changes requested):

kwok/Dockerfile
ifireball: @srivickynesh those defaults are for booting whole servers, you are not doing that here...

kwok/Dockerfile
ifireball: done

srivickynesh (approved):
LGTM

PR #121, opened 2023-11-22
chore(RHTAPWATCH-479): add member clusters to kwok

yftacherzog:
Changed the dockerfile so that multiple clusters are created.
Changed the inittab file so that all clusters are started and are proxied.
Changed the kwok yaml file to expose api ports for all clusters.

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide
instructions so we can reproduce. Please also list any relevant details
for your test configuration

# Checklist:

- [ ] I have performed a self-review of my code

### DIFF
--- a/kwok/Dockerfile
+++ b/kwok/Dockerfile
@@ -1,15 +1,22 @@
 FROM registry.k8s.io/kwok/cluster:v0.4.0-k8s.v1.27.3
 
-ARG CLUSTER_NAME=kwok
-ENV CLUSTER_NAME=${CLUSTER_NAME}
+ENV HOST_CLUSTER=host
+ENV MEMBER_CLUSTER=m01
+ENV RH_CLUSTER=rh01
 
 COPY --chmod=755 user-signups.sh toolchain.dev.openshift.com_usersignups.yaml ./
 COPY --chmod=644 inittab /etc/inittab
 
-RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
-    kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
-    sh ./user-signups.sh && \
-    rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
-    kwokctl stop cluster --name="$CLUSTER_NAME"
+RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$HOST_CLUSTER" || exit 1 && \
+    KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$MEMBER_CLUSTER" || exit 1 && \
+    KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$RH_CLUSTER" || exit 1 && \
+    kwokctl --name="$HOST_CLUSTER" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    CLUSTER_NAME="$HOST_CLUSTER" sh ./user-signups.sh && \
+    kwokctl --name="$MEMBER_CLUSTER" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    kwokctl --name="$RH_CLUSTER" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    kwokctl stop cluster --name="$HOST_CLUSTER" && \
+    kwokctl stop cluster --name="$MEMBER_CLUSTER" && \
+    kwokctl stop cluster --name="$RH_CLUSTER" && \
+    rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml
 
 ENTRYPOINT ["/sbin/init"]
--- a/kwok/inittab
+++ b/kwok/inittab
@@ -1,2 +1,6 @@
-::wait:/usr/local/bin/kwokctl start cluster --name kwok
-::respawn:/usr/local/bin/kwokctl --name kwok kubectl proxy --port=8080 --accept-hosts='^*$' --address=0.0.0.0
+::wait:/usr/local/bin/kwokctl start cluster --name host
+::wait:/usr/local/bin/kwokctl start cluster --name m01
+::wait:/usr/local/bin/kwokctl start cluster --name rh01
+::respawn:/usr/local/bin/kwokctl --name host kubectl proxy --port=8080 --accept-hosts='^*$' --address=0.0.0.0
+::respawn:/usr/local/bin/kwokctl --name m01 kubectl proxy --port=8070 --accept-hosts='^*$' --address=0.0.0.0
+::respawn:/usr/local/bin/kwokctl --name rh01 kubectl proxy --port=8060 --accept-hosts='^*$' --address=0.0.0.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,9 +21,9 @@ To run the Kwok container with the Kwok Kubernetes tool, follow these steps:
 
 1. Build the kwok container using the following command:
    ```
-   podman build --build-arg CLUSTER_NAME=kwok-cluster -t kwok -f kwok/Dockerfile .
+   podman build -t kwok -f kwok/Dockerfile kwok
    ```
-2. Bring the cluster up by running the following command from the 
+2. Bring the clusters up by running the following command from the 
    repo's root directory:
     ```
     podman kube play kwok/kwok_container_default.yml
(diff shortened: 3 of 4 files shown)


kwok/Dockerfile
```diff
@@ -1,15 +1,22 @@
 FROM registry.k8s.io/kwok/cluster:v0.4.0-k8s.v1.27.3
 
-ARG CLUSTER_NAME=kwok
-ENV CLUSTER_NAME=${CLUSTER_NAME}
+ENV CLUSTER_NAME=kwok
+ENV MEMBER1_CLUSTER_NAME=kwok-m1
+ENV MEMBER2_CLUSTER_NAME=kwok-m2
```
ifireball, line 5: I think we might as well call these "host", "m01" and "rh01" to make the names familiar to people who know about the RHTAP clusters.

Also I don't see a benefit for using env vars here as opposed to the explicit names.

kwok/Dockerfile
```diff
@@ -1,15 +1,22 @@
...
+    KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$MEMBER2_CLUSTER_NAME" || exit 1 && \
     kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    kwokctl --name="$MEMBER1_CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
+    kwokctl --name="$MEMBER2_CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
     sh ./user-signups.sh && \
     rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
-    kwokctl stop cluster --name="$CLUSTER_NAME"
+    kwokctl stop cluster --name="$CLUSTER_NAME" && \
+    kwokctl stop cluster --name="$MEMBER1_CLUSTER_NAME" && \
+    kwokctl stop cluster --name="$MEMBER2_CLUSTER_NAME"
```
ifireball, line 20: Perhaps dealing with each cluster in turn can reduce the probability of mistakes and failures.

kwok/inittab
```diff
@@ -1,2 +1,4 @@
-::wait:/usr/local/bin/kwokctl start cluster --name kwok
+::wait:/start-clusters.sh
```
ifireball, line 1: AFAIK you can have multiple "wait" statements, so no need of a script.

kwok/Dockerfile
yftacherzog: You mean multiple `RUN` lines?

kwok/start-clusters.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+/usr/local/bin/kwokctl start cluster --name kwok &
+/usr/local/bin/kwokctl start cluster --name kwok-m1 &
+/usr/local/bin/kwokctl start cluster --name kwok-m2
```
ifireball, line 4: the way this is written, the script may exit before "kwok" and "kwok-m1" finih starting if kwok-m2 finishes starting faster

kwok/inittab
yftacherzog: I started with that, and ran into some issues, but in hindsight they might have been user issues. I'll give it another try.

kwok/Dockerfile
ifireball: no, that would cause more image layers to be created, but ordering the commands within the single RUN statement differently.

kwok/start-clusters.sh
yftacherzog: Yes. This one is problematic. I'll see if I can make it work without the script.

kwok/inittab
yftacherzog: There seems to be no issue with multiple wait statements.
However, there seems to be an issue for me when starting a cluster after another - it seems to fail often.

The host is started in a second or so, then the first member is hanging until the timeout is reached and then the second member is starting in a second or so as well.

ifireball: so did using a script solve that somehow?

kwok/Dockerfile
yftacherzog: alright

kwok/inittab
yftacherzog: That's wasn't my original issue. The script did not have that issue, but I think the script is not the way to go. I think adding a short sleep before starting the next cluster addresses the issue, but I'm looking into that.

kwok/Dockerfile
yftacherzog: See whether you like it better now.

kwok/Dockerfile
yftacherzog: done

kwok/inittab
yftacherzog: I was not able to figure it out yet. It's working --- eventually. I'll try to explore a bit further. It might not happen today.

kwok/start-clusters.sh
yftacherzog: I removed the use of the script.

kwok/inittab
yftacherzog: We can introduce a script that will run the same startup command but will allow a few retries in case it fails. What do you think?

ifireball: I had a look at the set of running processes for a cluster:
```
  277 root      0:00 /root/.kwok/clusters/kwok/bin/etcd --name=node0 --auto-compaction-retention=1 --quota-backend-bytes=8589934592 --data-dir=/root/.kwok/clusters/kwok/etcd --initial-advertise-peer-urls=http://0.0.0.0:32766 --listen-pee
  278 root      0:02 /root/.kwok/clusters/kwok/bin/kube-apiserver --etcd-prefix=/registry --allow-privileged=true --etcd-servers=http://127.0.0.1:32765 --authorization-mode=Node,RBAC --bind-address=0.0.0.0 --secure-port=8080 --tls-cert-f
  279 root      0:00 /root/.kwok/clusters/kwok/bin/kwok-controller --manage-all-nodes=true --kubeconfig=/root/.kwok/clusters/kwok/kubeconfig.yaml --config=/root/.kwok/clusters/kwok/kwok.yaml --tls-cert-file=/root/.kwok/clusters/kwok/pki/
  280 root      0:00 /root/.kwok/clusters/kwok/bin/kube-controller-manager --node-monitor-period=10m0s --node-monitor-grace-period=1h0m0s --kubeconfig=/root/.kwok/clusters/kwok/kubeconfig.yaml --authorization-always-allow-paths=/healthz,
  281 root      0:00 /root/.kwok/clusters/kwok/bin/kube-scheduler --kubeconfig=/root/.kwok/clusters/kwok/kubeconfig.yaml --authorization-always-allow-paths=/healthz,/readyz,/livez,/metrics --bind-address=0.0.0.0 --secure-port=32762
```

This seems heavy, and it makes me think we have an issue with the resources in the container. I suspect we don't need all of that. 

Perhaps try disabling what you can in the `cluster create` command, we probably only need etcd and the API server.

yftacherzog: I'm able to remove `kube-controller-manager` and `kube-scheduler`, but it does not seem to affect the ability to get the container up. Yesterday, before the end of the day I could not make it fail, and this morning I had to run `podman system reset` to be able to make it start.

Also, waiting few seconds between starting the clusters doesn't make a difference. 

It's something about resources, but I'm not able to tell what it is.

Omeramsc (commented):
Can you elaborate further about the change in the commit message? Most of the additions are foreign to me.

yftacherzog:
> Can you elaborate further about the change in the commit message? Most of the additions are foreign to me.

I'll do it once I have it finalized. For now it's not working as expected yet.

ifireball (approved):
lgtm

kwok/Dockerfile
```diff
@@ -1,15 +1,22 @@
...
 
-RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$CLUSTER_NAME" || exit 1 && \
-    kwokctl --name="$CLUSTER_NAME" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
-    sh ./user-signups.sh && \
-    rm user-signups.sh toolchain.dev.openshift.com_usersignups.yaml && \
-    kwokctl stop cluster --name="$CLUSTER_NAME"
+RUN KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$HOST_CLUSTER" || exit 1 && \
+    KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$MEMBER_CLUSTER" || exit 1 && \
+    KWOK_KUBE_APISERVER_PORT=0 kwokctl create cluster --name "$RH_CLUSTER" || exit 1 && \
+    kwokctl --name="$HOST_CLUSTER" kubectl apply -f toolchain.dev.openshift.com_usersignups.yaml && \
```
Omeramsc, line 13: Does that mean we are applying the same workspace mapping to all 3 clusters? wouldn't that make it harder to test?

yftacherzog: line 13 is only applied to the host cluster.

Omeramsc: I referred to the fact all 3 lines use the same file (`toolchain.dev.openshift.com_usersignups.yaml`)

yftacherzog: In what sense are they using the same file?

yftacherzog: Not sure if that is ok or not.

Omeramsc: Maybe I'm misunderstanding the 'code', but all of them use `apply -f toolchain.dev.openshift.com_usersignups.yaml`

PR #141, opened 2023-12-18
chore(RHTAPWATCH-381): add tests for fetch-uj-records.sh

ShakedAviv50313:
implementing tests for fetch-uj-records.sh, currenly only for 5 out of 8 queries in querygen

[RHTAPWATCH-381](https://issues.redhat.com/browse/RHTAPWATCH-381)

## Type of change
Adding tests for scripts/fetch-uj-records.sh

# How Has This Been Tested?

Testing can be found at /fetch-uj-records/fetch-uj-records_test.go.
manually picked and loaded logs into a container with a local splunk server, then the logs were retrieved by the script running in the local environment.   

to activate the tests:
```
cd fetch-uj-records
go test
```
returns:
``` 
PASS
ok  	github.com/redhat-appstudio/segment-bridge.git/fetch-uj-records	0.809s
```


yftacherzog:
@ShakedAviv50313 unit tests are failing.

splunk/Dockerfile
```diff
@@ -30,6 +31,7 @@ RUN mkdir -p /opt/splunk/var/lib/splunk
 COPY --chmod=777 --from=builder /opt/splunk/etc/apps/search/local/indexes.conf /opt/splunk/etc/apps/search/local/indexes.conf
 COPY --chmod=777 --from=builder /var/splunk_buildtime_db/splunk/ /opt/splunk/var/lib/splunk/
 
+
```
Omeramsc, line 34: Nitpick, but as it has dragged for long, let's minimize the changes of this PR to what is absolutely necessary.

fetch-uj-records/fetch-uj-records_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/redhat-appstudio/segment-bridge.git/testfixture"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	scriptPath   = "../scripts/fetch-uj-records.sh"
+	filePathPass = "sample/fetch-uj-recordsPass"
+	filePathFail = "sample/fetch-uj-recordsFail"
```
Omeramsc, line 17: The Fail file is not included in this PR

ShakedAviv50313:
Details about the files changed in this PR:

1. .github/workflows/unit_tests.yaml - excluded this tests from the Ci's tests, the tests are running inside of a separate splunk container, therefore they are excluded from the CI tests.

2. fetch-uj-records/fetch-uj-recordsPass - the expected output of the script. (this is validated in the test with the script's actual output. the tests fails if it doesn't match).

3. fetch-uj-records/fetch-uj-records_test.go - The actual tests for fetch-uj-records.sh 

4. fetch-uj-records/splunk_container_default.yaml - same file as in splunk/splunk_container_default.yaml, golang "embed" does not support importing from other directories.

5. fetch-uj-records/splunk_container_template.tmpl - same as n.4

6.  splunk/Dockerfile - updated image, added added limits.conf and props.conf to load properly the logs into the splunk server in the container.

7. splunk/limits.conf - disabling some limits

8. splunk/props.conf - disabling truncate limit

9. splunk/splunk.go - added WithSplunk container, because in this file it is available in the "splunk" package

10. splunk/splunk_test.go - relates to n.10

11. splunk/tests/test_logs/fetch-uj-recordsFail.jsonl - logs which are loaded into the local splunk server (inside the testing container) but are not retrieved by fetch-uj-records.sh script

12. splunk/tests/test_logs/fetch-uj-recordsPass.jsonl -  logs which are loaded into the local splunk server (inside the testing container) and are retrieved by fetch-uj-records.sh script

13. splunk/tests/test_logs/log_example.log -deleted, there's no use of it.

fetch-uj-records/fetch-uj-records_test.go
```diff
@@ -0,0 +1,64 @@
...
+	scriptPath   = "../scripts/fetch-uj-records.sh"
+	filePathPass = "./fetch-uj-recordsPass"
+)
+
+func compareOutputs(output []byte, filePath string) bool {
+	/*Recieves the script's output and compares it
+	with a file containing the expected output*/
+	fileOutput, err := os.ReadFile(filePath)
+	if err != nil {
+		fmt.Printf("error: %s\n", err.Error())
```
Omeramsc, line 27: Use `t *testing.T` and t.Log instead of Printf
Apply this to all other Printf in the test files.

fetch-uj-records/fetch-uj-records_test.go
```diff
@@ -0,0 +1,64 @@
...
+		return false
+	}
+	return bytes.Equal(fileOutput, output)
+}
+
+func runAndValidateScript(t *testing.T, filePath, scriptPath string) bool {
+	/*Returns true if a given scripts runs successfully
+	and it's output is equal to a given file's content*/
```
Omeramsc, line 35: I think the test should be self-explanatory without docstrings. if you don't think it is clear enough, you should how to make it more readable, rather than adding explanation. 
Apply this to all other docstrings in the test files.

fetch-uj-records/fetch-uj-records_test.go
```diff
@@ -0,0 +1,64 @@
...
+	require.NoError(t, os.Setenv("SPLUNK_APP_NAME", "-"), "Failed to set SPLUNK_API_URL")
+	require.NoError(t, os.Setenv("SPLUNK_INDEX", "test_index"), "Failed to set SPLUNK_INDEX")
+	require.NoError(t, os.Setenv("QUERY_EARLIEST_TIME", "2023-01-01T10:46:34.955712833Z"), "Failed to set QUERY_EARLIEST_TIME")
+	splunk.WithSplunkContainer(t, func(deployment containerfixture.FixtureInfo) {
+		splunkAppApiURL := splunk.GetSplunkAppAPIEndpoint("localhost", deployment.ApiPort, "nobody", "-")
+		require.NoError(t, os.Setenv("SPLUNK_API_URL", "localhost:"+deployment.ApiPort), "Failed to set SPLUNK_API_URL")
+		netrcPath, err := containerfixture.GetNetrcPath()
+		if err != nil {
+			fmt.Printf("error: %s\n", err.Error())
+			return
```
Omeramsc, line 56: why not assert here \ fail the test in that case? we won't be able to connect to Splunk without a netrc.

fetch-uj-records/fetch-uj-records_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func runAndValidateScript(t *testing.T, filePath, scriptPath string) bool {
+	/*Returns true if a given scripts runs successfully
+	and it's output is equal to a given file's content*/
+	cmd := exec.Command(scriptPath)
+	output, err := cmd.Output()
+	if err != nil {
+		fmt.Printf("error: %s\n", err.Error())
+		return false
```
Omeramsc, line 40: why not assert\ fail the test here?

fetch-uj-records/splunk_container_default.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
Omeramsc, line 1: Why do we copy the `splunk_container_default.yaml` here instead of using the two templates available in splunk? 
take a look at `uid_map` and `ws_map`, they both use the kwok container template, but without duplicating it from the `kwok` directory.

fetch-uj-records/splunk_container_template.tmpl
```diff
@@ -0,0 +1,21 @@
+---
```
Omeramsc, line 1: same

splunk/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/splunk/splunk@sha256:7eca42b8a7a777053e1d0420287a683167ed228ab09444b35b3f4a8f54cb6cd4 AS builder
+FROM docker.io/splunk/splunk@sha256:20b1376a1677ef6eff611b1d1e25e3a170a715c646c2c27711bbb65b03f78ede AS builder
```
Omeramsc, line 1: Is this change needed, or can we save it for another PR?

splunk/splunk.go
```diff
@@ -17,3 +21,19 @@ func GetSplunkAppAPIEndpoint(host, port, ownerName, appName string) string {
...
 	return fmt.Sprintf("%s/search/v2/jobs/export", appAPIEndpoint)
 }
+
+//go:embed splunk_container_template.tmpl
+var splunkServiceManifest string
+
+const NotUpErrorMsg = "The %s instance is not up, cannot verify indexing for tests."
+
+// WithSplunkContainer builds a splunk container
+func WithSplunkContainer(t *testing.T, testFunc func(containerfixture.FixtureInfo)) {
```
Omeramsc, line 31: This is a test fixture, why move it into the Splunk module?

splunk/splunk.go
```diff
@@ -17,3 +21,19 @@ func GetSplunkAppAPIEndpoint(host, port, ownerName, appName string) string {
 func GetSearchAPIEndpoint(appAPIEndpoint string) string {
 	return fmt.Sprintf("%s/search/v2/jobs/export", appAPIEndpoint)
 }
+
+//go:embed splunk_container_template.tmpl
+var splunkServiceManifest string
+
+const NotUpErrorMsg = "The %s instance is not up, cannot verify indexing for tests."
+
+// WithSplunkContainer builds a splunk container
```
Omeramsc, line 30: While it may need better documentation, I don't believe this line gives any more context than the function does by reading it.

Omeramsc (changes requested):
Good progress, I left a few comments

splunk/Dockerfile
ShakedAviv50313: we can save it for another PR, Eran told me that this image should be upgraded.

splunk/splunk.go
ShakedAviv50313: Because i'm using it also in the fetch-uj-records_test.go and I couldn't import it if it wasn't in the module

Omeramsc (approved):
LGTM, good job

yftacherzog (commented):
@ShakedAviv50313 you mentioned that those tests are not expected to pass, but I think we still need CI to pass. What do we do about that? Do we have similar tests that should fail for similar reasons? How do we handle those?

fetch-uj-records/fetch-uj-recordsPass
yftacherzog, line 1: The name of this file is very similar to the name of the input file. I think they should be renamed so it's clear what each one is about.

splunk/tests/test_logs/fetch-uj-recordsFail.jsonl
yftacherzog, line 1: Are we using the content of this file. Is there a pass/fail distinction for the logs?

ShakedAviv50313: These logs are loaded into the splunk server and should not be retrieved by any query in querygen.
We need this file to make sure the script returns only the correct logs. the logs returned and the logs not returned are separated into two files so it would be clearer

ShakedAviv50313:
``` @ShakedAviv50313 you mentioned that those tests are not expected to pass, but I think we still need CI to pass. What do we do about that? Do we have similar tests that should fail for similar reasons? How do we handle those? ```

@yftacherzog 
The tests run inside of a splunk container which is not the CI container, of course I require the tests to pass, on the CI the tests wouldn't pass because of the container so it's excluded from the CI's tests, the CI now passes.

Omeramsc:
> ed that those tests are not expected to pass, but I think we still need CI to pass. What do we do about that? Do we have similar tests that shoul

I'll clarify: Currently it had been decided that any test that require running a container in order to perform the test (splunk, uid_map, ws_map, kwok, and this one) should be excluded from our CI, while ensuring they pass locally. No need for "dummy tests" as they won't provide us with verification for the results we expect to see.
That being said, the tests which are not excluded from the unit-test CI should pass (RHTAP is expected to fail currently, can be ignored).

fetch-uj-records/fetch-uj-recordsPass
ShakedAviv50313: Changed the name to requiredOutput

splunk/tests/test_logs/fetch-uj-recordsFail.jsonl
yftacherzog: ok, so maybe just rename it, because we have no failure scenarios here. recordsToBeIgnored maybe?

ShakedAviv50313: Done, changed to recordsNotFetched

yftacherzog (approved):
lgtm

PR #147, opened 2024-01-04
chore(RHTAPWATCH-160): Run SAST checks for each opened PR

klakshma21:
RHTAPWATCH-160

## Type of change
- Add SAST check

Output:

image

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,9 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.25.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+exclude:
+  global:
+    - db.go
+    - 'kwok/host-operator/'
+    
+patch: {}
--- a/.tekton/segment-bridge-pull-request.yaml
+++ b/.tekton/segment-bridge-pull-request.yaml
@@ -30,6 +30,10 @@ spec:
     value: .
   - name: revision
     value: '{{revision}}'
+  - default: "snyk-secret"
+    description: Snyk Token Secret Name
+    name: snyk-secret
+    type: string
   pipelineSpec:
     finally:
     - name: show-sbom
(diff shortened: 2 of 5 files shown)


klakshma21:
@amisstea @ifireball @hmariset The SAST check now passed after excluding the filePathPass related files.
image

Kousalya1998:
@amisstea We can merge this PR as it is up to date. From the next PR I will use my old account

hmariset:
Based on @Kousalya1998 request, I'm reopening this PR again

.snyk
```diff
@@ -0,0 +1,11 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.25.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+exclude:
+  global:
```
amisstea, line 5: It's not obvious to me why we need to exclude these sources. Please add comments explaining why with a few details about the root problem.

Also, we shouldn't need the wildcard pattern matching on the last two. I expect something like this should work (untested):
`**/fetch-uj-records/fetch-uj-records_test.go**` -> `fetch-uj-records/fetch-uj-records_test.go`

klakshma21: I will leave a comment there. I excluded them because it threw an error for using hardcoded passwords for having filePathPass in those files. @amisstea

amisstea: For anyone else following, snyk disagrees with our choice of naming for the `filePathPass` variables in our test files. Simply renaming those variables resolved the errors.

amisstea (approved):

amisstea (changes requested):
@klakshma21 please run go fmt to resolve the pre-commit errors.

klakshma21:
/

> @klakshma21 please run go fmt to resolve the pre-commit errors.

@amisstea Corrected the indentation errors.

amisstea (approved):

